PR #1, opened 2018-01-17
Common components refactor

akinwale:


.eslintrc.json
```diff
@@ -0,0 +1,30 @@
+{
```
kauffj: It's a dream to have linting and prettier configuration files shared across JS repos, right?

kauffj: (By dream I mean unrealistic)

build/index.js
```diff
@@ -1,5 +1,14 @@
-module.exports =
```
kauffj: This shouldn't be checked in.

package.json
```diff
@@ -17,6 +17,7 @@
   "main": "build/index.js",
   "scripts": {
     "build": "webpack",
+    "precommit": "lint-staged",
```
kauffj: Please remove any unused depdencies.

src/index.js
```diff
@@ -1,4 +1,154 @@
...
+  selectIsFetchingTransactions,
+  selectIsSendingSupport,
+  selectReceiveAddress,
+  selectGettingNewAddress,
+  selectDraftTransaction,
+  selectDraftTransactionAmount,
+  selectDraftTransactionAddress,
+  selectDraftTransactionError,
+  selectBlocks,
+} from 'redux/selectors/wallet';
```
kauffj: What's the purpose of this file?

src/lbry.js
```diff
@@ -12,19 +12,6 @@ function apiCall(method, params, resolve, reject) {
   return jsonrpc.call(Lbry.daemonConnectionString, method, params, resolve, reject, reject);
```
kauffj: jsonrpc logic can be moved into this file

kauffj: Also, all of the properties on `Lbry` can be killed and `Lbry` can just be the API proxy.

src/lbryapi.js
```diff
@@ -1,7 +1,7 @@
-import Lbry from 'lbry';
 import querystring from 'querystring';
 
-const LbryApi = {
+const Lbryapi = {
```
kauffj: I think we should consider naming this something else. If a new user is using this repo to build something, it's confusing to have two objects, `Lbry`, and `LbryApi`, and then first one is the LBRY API!

Maybe call this `LbryIncApi` or `LbryIoApi`? Not in love with either.

src/redux/actions/app.js
```diff
@@ -0,0 +1,24 @@
+import * as ACTIONS from 'constants/action_types';
+
+export function doOpenModal(modal, modalProps = {}) {
```
kauffj: All of these actions should probably be outside of `lbry-redux`, eh?

src/redux/actions/claims.js
```diff
@@ -0,0 +1,109 @@
+import * as ACTIONS from 'constants/action_types';
+import * as MODALS from 'constants/modal_types';
+import Lbry from 'lbry';
+import { buildURI, normalizeURI } from 'lbryURI';
+import { doOpenModal } from 'redux/actions/app';
```
kauffj: All of these will need to be removed from other actions.

src/redux/actions/navigation.js
```diff
@@ -0,0 +1,64 @@
+import * as ACTIONS from 'constants/action_types';
```
kauffj: Should these actions be in `lbry-redux`? I'm inclined to say no.

build/index.js
akinwale: This was generated by webpack. I'll see if I can remove it.

src/index.js
akinwale: The exports exist to expose names that need to be used from the source file, so that they can be included in the webpack build, which makes them accessible when the lbry-redux package is being used elsewhere. This is similar to how other React packages work. For example, https://github.com/reactjs/react-redux/blob/master/src/index.js

build/index.js is the webpack output.

src/lbryapi.js
akinwale: LbryWebApi or Lbrydotio, maybe?

src/lbry.js
```diff
@@ -159,6 +204,7 @@ Lbry.publishDeprecated = (params, fileListedCallback, publishedCallback, errorCa
     { once: true }
   );
 
+  // eslint-disable-next-line no-use-before-define
   lbryProxy.publish(params).then(
```
seanyesmunt: All of the `lbry.publish` code was removed. The weird stuff with pending publishes was removed from this file and is handled inside the `publish` reducer

src/lbryapi.js
seanyesmunt: I like `LbryIncApi`.

Whatever works though, I agree it might be nicer to have a larger difference between this and `lbry`

src/redux/actions/app.js
seanyesmunt: Did we decide on something like `DO_NOTIFY`?

src/redux/actions/wallet.js
```diff
@@ -0,0 +1,176 @@
...
+import { doOpenModal, doShowSnackBar } from 'redux/actions/app';
+import {
+  selectBalance,
+  selectDraftTransaction,
+  selectDraftTransactionAmount,
+} from 'redux/selectors/wallet';
+
+export function doUpdateBalance() {
+  return dispatch => {
+    Lbry.wallet_balance().then(balance =>
```
seanyesmunt: https://github.com/lbryio/lbry-app/blob/master/src/renderer/redux/actions/wallet.js#L12

akinwale: > Did we decide on something like `DO_NOTIFY`?
Yes, we did. I was supposed to delete the `src/redux/actions/app.js` file. It's not used anywhere in `lbry-redux` at this point.

You can see the notifications related code in `src/redux/actions/notification.js`, `src/redux/reducers/notification.js` and `src/redux/selectors/notification.js`. We'll need to update `lbry-app` to make use of the new action, which will be the next step after the merge is completed.

PR #7, opened 2018-04-19
Simplify Media Identification Function

ProfessorDey:
Rebased from LBRY/lbry-app PR# 1304 with corrections requested by Codacy (Though may still complain regarding spacing).

This just makes the code for finding documents a lot easier. the regex patterns and the media type are converted into keypairs, [regex, mediatype] which is then reduced to a single value, which will either be the original filename if there are no matching extensions, or the mediatype. So long as the mediatype does not contain an extension at the end of the string, this system should work far more simply.

This will also allow for easier addition of new extensions to the function, so long as they all follow the same pattern for the regex:
^ = Start of string
.+ = 1 or more characters of any type
. = A literal dot
(a|b|c) = A list of ORed extensions within a capture group '( )'
$ = End of string, so it is certain this is the actual extension of the filename

Won't be hard to add things like Archives, Executables, etc with just a single extra line now.

### DIFF
--- a/src/lbry.js
+++ b/src/lbry.js
@@ -112,20 +112,16 @@ Lbry.getMediaType = (contentType, fileName) => {
   if (contentType) {
     return /^[^/]+/.exec(contentType)[0];
   } else if (fileName) {
-    const dotIndex = fileName.lastIndexOf('.');
-    if (dotIndex === -1) {
-      return 'unknown';
-    }
-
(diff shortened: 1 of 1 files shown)


tzarebczan:
@seanyesmunt please review when you get a moment. See my previous comments on https://github.com/lbryio/lbry-app/pull/1304#issuecomment-381777069, let me know if we need to discuss.

PR #17, opened 2018-05-05
Update modal types constants and export them to use on lbry-app.

dan1d:


seanyesmunt:
Please run `yarn build` then commit and push. We need the build to be on github so the app can consume this module

dan1d:
Hi @seanyesmunt given that I'm new to lbry I thought someone else would actually build the new version of the bundle.js file, if you allow me to do that, I can inject code that can compromise the security of lbry-app as I can obfuscate a package inside the bundle and you wouldn't know, I won't do it but that's a possibility that I shouldn't have. 
Just my 2 cents

seanyesmunt:
@dan1d That is a good point and something we will discuss moving forward. Thanks

PR #19, opened 2018-05-09
Add modal-type for external links

btzr-io:
### Confirmation for external links: https://github.com/lbryio/lbry-app/issues/1442

Required by https://github.com/lbryio/lbry-app/pull/1441

### DIFF
--- a/src/types/Notification.js
+++ b/src/types/Notification.js
@@ -15,5 +15,4 @@ export type Notification = {
 // Used for retreiving data from redux store
 export type NotificationProps = {
   uri: ?string,
-  url: ?string,
 };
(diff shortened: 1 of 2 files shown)


src/types/Notification.js
```diff
@@ -15,4 +15,5 @@ export type Notification = {
 // Used for retreiving data from redux store
 export type NotificationProps = {
   uri: ?string,
+  url: ?string,
```
btzr-io: @seanyesmunt I need to pass this data to the modal, is there a better way to handle this?

seanyesmunt: why can't this be the uri prop?

btzr-io: Ok I guess I could use that, but it's not the `uri` from a claim,  it could be any url like: `https://www.patreon.com/`

PR #20, opened 2018-05-09
mark support transaction as completed

daovist:
While investigating the 0LBC doSendSupport issue (lbry-app #738) I noticed that if I sent a tip, I had to restart the app to send another, because this wasn't being dispatched.


tzarebczan:
That must fix https://github.com/lbryio/lbry-app/issues/1398

PR #27, opened 2018-05-21
Add function to convert an URI to  an url that can be shared and opened with lbry-app

dan1d:
This function will parse a lbry URI and convert it to an https protocol, Let me know if this is the correct place to add it, we may need to use it on lbry-android sometime(tap on a video to share an url or something like that)
This PR is needed for [PR on lbry-app#1486](https://github.com/lbryio/lbry-app/pull/1486)
So for example given this URI:
`lbry://why-doesn-t-capitalism-work-for#36cd5b217bcc4083a6fc35cc4de2cc1196581ef9`
changes to:
`https://open.lbry.io/why-doesn-t-capitalism-work-for#36cd5b217bcc4083a6fc35cc4de2cc1196581ef9`

### DIFF
--- a/src/lbryURI.js
+++ b/src/lbryURI.js
@@ -230,12 +230,5 @@ export function isURIClaimable(URI) {
 }
 
 export function convertToShareLink(URI) {
-  if (URI.match(/pending_claim/)) return URI;
-
-  const { claimName, path, bidPosition, claimSequence, claimId } = parseURI(URI);
-  return buildURI(
-    { claimName, path, claimSequence, bidPosition, claimId },
-    true,
-    'https://open.lbry.io/'
-  );
+  return buildURI(parseURI(URI), true, 'https://open.lbry.io/');
 }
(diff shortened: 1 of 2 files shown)


tzarebczan:
@dan1d nice touch, this could be super useful to add as a share LBRY URL button in the app!

seanyesmunt:
@skhameneh

dan1d:
@tzarebczan thanks, let me know if you may want that button I'm happy to work on it!

src/lbryURI.js
```diff
@@ -228,3 +228,14 @@ export function isURIClaimable(URI) {
     !parts.path
   );
 }
+
+export function convertToShareLink(URI) {
+  if (URI.match(/pending_claim/)) return URI;
+
+  const { claimName, path, bidPosition, claimSequence, claimId } = parseURI(URI);
```
skhameneh: Do you need these split out?
Seems like you can just do:
```
+  return buildURI(
+    parseURI(URI),
+    true,
+    'https://open.lbry.io/'
+  );
```

dan1d:
Code updated @skhameneh

PR #32, opened 2018-05-25
filter search suggestions so the search query isn't suggested twice

seanyesmunt:
If the search suggestion api returns a query that is the same as the one the user typed in the search bar, we should remove it so there aren't two identical search suggestions.

Fixes https://github.com/lbryio/lbry-app/issues/1495

### DIFF
--- a/src/redux/actions/search.js
+++ b/src/redux/actions/search.js
@@ -107,12 +107,6 @@ export const getSearchSuggestions = (value: string) => dispatch => {
         type: SEARCH_TYPES.SEARCH,
       }
     );
-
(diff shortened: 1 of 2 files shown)


src/redux/actions/search.js
```diff
@@ -107,12 +107,6 @@ export const getSearchSuggestions = (value: string) => dispatch => {
         type: SEARCH_TYPES.SEARCH,
       }
     );
-
-    // If it's a valid url, don't fetch any extra search results
-    dispatch({
```
seanyesmunt: We were dispatching twice with the same data. Don't need this

PR #37, opened 2018-05-30
remove notification props

daovist:
this was getting in the way of other data I needed to pass to a notification and will continue to be an issue for other use cases


src/redux/actions/notifications.js
```diff
@@ -2,7 +2,7 @@
 import * as ACTIONS from 'constants/action_types';
 import Notification from 'types/Notification';
 
-export function doNotify(notification: Notification, notificationProps: NotificationProps) {
+export function doNotify(notification: Notification, notificationProps) {
```
seanyesmunt: We can keep still keep this type here. I would add your new value to it.

PR #38, opened 2018-05-31
Fix convertToShareLink

tzarebczan:
Pass correct values to builduri. Otherwise, you get this when right clicking something:
```
Uncaught Error: Received a channel content URI, but claim name and channelName do not match. "name" represents the value in the name position of the URI (lbry://name...), which for channel content will be the channel name. In most cases, to construct a channel URI you should just pass channelName and contentName.
    at buildURI (C:\Program Files\LBRY\resources\app.asar\node_modules\lbry-redux\dist\bundle.js:126
```

@dan1d did you run into this during your testing?

### DIFF
--- a/dist/bundle.js
+++ b/dist/bundle.js
@@ -1319,7 +1319,14 @@ function isURIClaimable(URI) {
 }
 
 function convertToShareLink(URI) {
-  return buildURI(parseURI(URI), true, 'https://open.lbry.io/');
+  var _parseURI2 = parseURI(URI),
+      claimName = _parseURI2.claimName,
+      path = _parseURI2.path,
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Is it just that there are more values than that in the `parseURI` which causes the `buildURI` to fail?

tzarebczan:
Believe so. I copied the pattern from another function in the same file. I see builduri being called from lbry app with claim name and claim ID typically.

dan1d:
@tzarebczan It worked for me locally before the requested review change, after that I changed the code I think that I didn't run the command `yarn build` to actually test the new change, so the app still worked for me.

Does your commit fix the issue ?

PR #44, opened 2018-06-05
add methods related to fetching trending content

akinwale:


src/redux/actions/claims.js
```diff
@@ -161,6 +161,40 @@ export function doFetchFeaturedUris() {
...
+        },
+      ];
+      dispatch(batchActions(...actions));
+    };
+
+    const failure = () => {
+      dispatch({
+        type: ACTIONS.FETCH_TRENDING_CONTENT_COMPLETED,
+        data: {
+          uris: {},
```
seanyesmunt: Should this be an array? It's an array onSuccess

akinwale: @seanyesmunt Indeed, it should be. Pushed a new commit.

PR #56, opened 2018-07-13
Remove getMediaType function

btzr-io:
Currently we use and call this method inside the `lbry-desktop` repository: 
https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/util/getMediaType.js

### DIFF
--- a/src/lbry.js
+++ b/src/lbry.js
@@ -109,29 +109,6 @@ Lbry.connect = () => {
   return Lbry.connectPromise;
 };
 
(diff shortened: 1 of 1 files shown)


seanyesmunt:
I guess we should move that file into `lbry-redux`

It is used [here](https://github.com/lbryio/lbry-android/blob/afaa00d635b29c8d6ccfe98c05a52e76f6a22561/app/src/page/file/view.js#L207)

Issue #36, opened 2018-05-29
Add wallet encryption support (selectors, actions, and reducers)


skhameneh:
## The Issue
Per https://github.com/lbryio/lbry-app/issues/762 and https://github.com/lbryio/lbry-app/issues/1097, wallet encryption support (selectors, actions, and reducers) needs to be added

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
@skhameneh how's it coming along?

skhameneh:
Proceeding, will have it added this week. 😄

PR #59, opened 2018-07-18
Add encryption flow to redux

skhameneh:
See lbryio/lbry-redux#36

lbryio/lbry-desktop#1097 will be updated tomorrow with a dependency on these changes.


src/redux/selectors/wallet.js
```diff
@@ -61,7 +95,7 @@ export const selectTransactionItems = createSelector(selectTransactionsById, byI
           txid,
           date: tx.timestamp ? new Date(Number(tx.timestamp) * 1000) : null,
           amount,
-          fee: amount < 0 ? (-1 * tx.fee) / append.length : 0,
+          fee: amount < 0 ? -1 * tx.fee / append.length : 0,
```
skhameneh: This should not be here, I will check my local commits and try to find why this change is appearing.

seanyesmunt: Guessing it's from Prettier.

src/redux/actions/wallet.js
```diff
@@ -242,3 +242,105 @@ export function doSendSupport(amount, claimId, uri, successCallback, errorCallba
...
+  };
+}
+
+export function doWalletLock() {
+  return dispatch => {
+    dispatch({
+      type: ACTIONS.WALLET_LOCK_START,
+    });
+
+    Lbry.wallet_lock().then(() => {
```
seanyesmunt: What if this command fails?

src/redux/reducers/wallet.js
```diff
@@ -14,6 +14,13 @@ const defaultState = {
   gettingNewAddress: false,
   draftTransaction: buildDraftTransaction(),
   sendingSupport: false,
+  walletIsEncrypted: false,
+  walletEncryptPending: false,
+  walletEncryptSucceded: null,
+  walletDecryptPending: false,
+  walletDecryptSucceded: null,
+  walletUnlockPending: false,
+  walletUnlockSucceded: null,
```
seanyesmunt: There should probably be an error attribute here too. Not sure what the daemon returns but it would be nice to know if there was an error decrypting or some unrelated error.

src/redux/selectors/wallet.js
```diff
@@ -2,6 +2,40 @@ import { createSelector } from 'reselect';
 
 export const selectState = state => state.wallet || {};
 
+export const selectWalletState = selectState;
+
+export const selectWalletIsEncrypted = state => selectWalletState(state).walletIsEncrypted === true;
```
seanyesmunt: I would use `createSelector` here like you did below.

src/redux/reducers/wallet.js
```diff
@@ -131,14 +139,106 @@ reducers[ACTIONS.SUPPORT_TRANSACTION_FAILED] = (state, action) =>
...
+    walletLockResult: action.result,
+  });
+
+reducers[ACTIONS.WALLET_LOCK_FAILED] = (state, action) =>
+  Object.assign({}, state, {
+    walletLockPending: false,
+    walletLockSucceded: false,
+    walletLockResult: action.result,
+  });
+
```
seanyesmunt: Since this won't get merged in the immediate future. We should add Flow types to this.

src/redux/reducers/wallet.js
```diff
@@ -6,6 +7,36 @@ const buildDraftTransaction = () => ({
   address: undefined,
 });
 
+// TODO: Split into common success and failure types
+// See details in https://github.com/lbryio/lbry/issues/1307
+type ActionResult = {
```
seanyesmunt: Maybe `EncryptActionResult`?

skhameneh: Likely it'll split into multiple at some point.
We don't have any predictable pattern at the moment and this only contains `result` and could apply to any other action, I can change it but I still think we'll end up with another common pattern.

seanyesmunt: 👍

PR #64, opened 2018-08-02
User history

daovist:
See [lbryio/lbry-desktop/issue/1009](https://github.com/lbryio/lbry-desktop/issues/1009)

- [x] add actions for recording history
- [x] rename position tracking actions
- [x] add `UserHistoryPage` to `My LBRY` in sidebar
- [x] fix flow errors


src/constants/action_types.js
```diff
@@ -80,11 +80,15 @@ export const CREATE_CHANNEL_COMPLETED = 'CREATE_CHANNEL_COMPLETED';
 export const PUBLISH_STARTED = 'PUBLISH_STARTED';
 export const PUBLISH_COMPLETED = 'PUBLISH_COMPLETED';
 export const PUBLISH_FAILED = 'PUBLISH_FAILED';
-export const SET_PLAYING_URI = 'PLAY_URI';
 export const FETCH_BLACK_LISTED_CONTENT_STARTED = 'FETCH_BLACK_LISTED_CONTENT_STARTED';
 export const FETCH_BLACK_LISTED_CONTENT_COMPLETED = 'FETCH_BLACK_LISTED_CONTENT_COMPLETED';
 export const FETCH_BLACK_LISTED_CONTENT_FAILED = 'FETCH_BLACK_LISTED_CONTENT_FAILED';
 export const BLACK_LISTED_CONTENT_SUBSCRIBE = 'BLACK_LISTED_CONTENT_SUBSCRIBE';
+export const SET_PLAYING_URI = 'SET_PLAYING_URI';
```
seanyesmunt: Do we need `SET_PLAYING_URI` if we have `MEDIA_PLAY`? Can that be used to set the same values?

daovist: I forgot to remove `MEDIA_PLAY` from this repo. I'll do so now and push...

src/redux/selectors/navigation.js
```diff
@@ -62,7 +62,8 @@ export const selectNavLinks = createSelector(
       page === 'invite' ||
       page === 'backup';
 
-    const isMyLbryPage = page => page === 'downloaded' || page === 'published';
+    const isMyLbryPage = page =>
+      page === 'downloaded' || page === 'published' || page === 'user_history';
```
kauffj: Aren't these pages app-specific and thus inappropriate for lbry-redux?

Even if is redux appropriate, this is a dangerous/fragile pattern - how would someone changing the name of a page or route know to update this function?

seanyesmunt: Yeah... This selector should probably live in the app.

daovist: Agreed.

seanyesmunt:
Moved `selectNavLinks` into the app

PR #67, opened 2018-08-14
Move search suggestion logic to selector

seanyesmunt:
### Changes
- Default first suggestion to be a search
- Most of the search suggestion logic is now in a selector
- Just pass in the api results to the search reducer with a `query` key so previous suggestions are preserved when using the backspace/retyping the same query

### DIFF
--- a/src/redux/selectors/search.js
+++ b/src/redux/selectors/search.js
@@ -1,9 +1,13 @@
+import * as SEARCH_TYPES from 'constants/search';
+import { normalizeURI, parseURI } from 'lbryURI';
 import { selectCurrentPage, selectCurrentParams } from 'redux/selectors/navigation';
 import { createSelector } from 'reselect';
 
 export const selectState = state => state.search || {};
 
 export const selectSearchValue = createSelector(selectState, state => state.searchQuery);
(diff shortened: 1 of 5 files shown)


src/redux/selectors/search.js
```diff
@@ -37,3 +41,79 @@ export const selectWunderBarAddress = createSelector(
 );
 
 export const selectSearchBarFocused = createSelector(selectState, state => state.focused);
+// export const selectSear
+
+export const selectSearchSuggestions = createSelector(
```
akinwale: Just an idea I had. I was thinking we should have an additional selector called `selectLocalSearchSuggestions` which could possibly be used in cases where an API call doesn't necessarily have to be made.

seanyesmunt: Maybe based on previous searches?

seanyesmunt:
@akinwale I'll create a PR to lbry-android after adding this into the app. The changes should be very minor.

src/redux/selectors/search.js
akinwale: Yeah, that makes sense.

PR #68, opened 2018-08-15
handle new daemon status response

seanyesmunt:
Changes for 0.21.0 daemon


src/redux/actions/wallet.js
```diff
@@ -337,16 +337,11 @@ export function doWalletStatus() {
...
-      if (result && !result.error) {
+    Lbry.status().then(status => {
+      if (status && status.wallet) {
         dispatch({
           type: ACTIONS.WALLET_STATUS_COMPLETED,
-          result,
-        });
-      } else {
-        dispatch({
-          type: ACTIONS.WALLET_STATUS_FAILED,
```
seanyesmunt: The `WALLET_STATUS_FAILED` action type wasn't being used anywhere.

seanyesmunt:
Merging this now so `master` will work with the desktop app. @akinwale @skhameneh I would appreciate a review when either of you can.

src/redux/actions/wallet.js
```diff
@@ -337,16 +337,11 @@ export function doWalletStatus() {
...
-          result,
+          result: status.wallet.is_encrypted,
```
akinwale: Changing the result to return just the `is_encrypted` value sort of makes this confusing because the function is named `doWalletStatus`. Maybe the name of the function could be changed to something more meaningful in this instance.

PR #73, opened 2018-08-28
encode # on search

seanyesmunt:


tzarebczan:
Looks good here:  https://lighthouse.lbry.io/search?s=GameTales%20%2311:%20NES%20The%20console%20that%20beat%20communism,%20The%20best%20console%20ever%20made&size=20&from=0

PR #78, opened 2018-09-13
In-App notifications (redux logic)

btzr-io:
### WIP
See: https://github.com/lbryio/lbry-desktop/issues/695


src/redux/reducers/notifications.js
```diff
@@ -39,6 +63,25 @@ reducers[ACTIONS.DISMISS_NOTIFICATION] = state => {
...
+  const { index } = action.data;
+  stack.splice(index, 1);
+
+  return Object.assign({}, state, {
+    stack,
+  });
+};
+
+reducers[ACTIONS.CLEAR_NOTIFICATIONS] = state => {
+  const stack = Object.assign([], state.stack);
```
btzr-io: Is this necessary?

btzr-io: can it be just `const stack = [];` ?

btzr-io:
Ready for review ^^

seanyesmunt:
Will give this a more in-depth review, but is there any way we could add this to the existing notification stack? We would probably need additional values like the time/is_read but I think that would be preferable over keeping it separate.

PR #79, opened 2018-09-24
Support for lbrynet daemon v0.30.0

eukreign:
Ready for review

### DIFF
--- a/src/redux/actions/wallet.js
+++ b/src/redux/actions/wallet.js
@@ -2,13 +2,14 @@ import * as ACTIONS from 'constants/action_types';
 import Lbry from 'lbry';
 import { doNotify } from 'redux/actions/notifications';
 import { selectBalance } from 'redux/selectors/wallet';
+import { creditsToString } from 'util/formatCredits';
 
 export function doUpdateBalance() {
   return (dispatch, getState) => {
     const {
       wallet: { balance: balanceInStore },
     } = getState();
-    Lbry.wallet_balance().then(balance => {
+    Lbry.account_balance().then(balance => {
       if (balanceInStore !== balance) {
         dispatch({
           type: ACTIONS.UPDATE_BALANCE,
@@ -63,7 +64,7 @@ export function doGetNewAddress() {
     });
 
     // Removed localStorage use, since address is expected to be stored in redux store
-    Lbry.wallet_new_address().then(address => {
+    Lbry.address_unused().then(address => {
       dispatch({
         type: ACTIONS.GET_NEW_ADDRESS_COMPLETED,
         data: { address },
@@ -78,7 +79,7 @@ export function doCheckAddressIsMine(address) {
       type: ACTIONS.CHECK_ADDRESS_IS_MINE_STARTED,
     });
 
-    Lbry.wallet_is_address_mine({ address }).then(isMine => {
+    Lbry.address_is_mine({ address }).then(isMine => {
       if (!isMine) dispatch(doGetNewAddress());
 
       dispatch({
@@ -109,8 +110,8 @@ export function doSendDraftTransaction(address, amount) {
       type: ACTIONS.SEND_TRANSACTION_STARTED,
     });
 
-    const successCallback = results => {
-      if (results === true) {
+    const successCallback = response => {
+      if (response.txid) {
         dispatch({
           type: ACTIONS.SEND_TRANSACTION_COMPLETED,
         });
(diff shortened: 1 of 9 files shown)


src/redux/reducers/claims.js
```diff
@@ -96,7 +96,7 @@ reducers[ACTIONS.FETCH_CHANNEL_LIST_COMPLETED] = (state, action) => {
 
   claims.forEach(claim => {
     myChannelClaims.add(claim.claim_id);
-    byId[claims.claim_id] = claim;
+    byId[claim.claim_id] = claim;
```
seanyesmunt: This never worked, not sure how the channel selector worked in the first place ¯\_(ツ)_/¯

seanyesmunt: Appreciate it ¯\\\_(ツ)_/¯

src/redux/actions/wallet.js
```diff
@@ -109,8 +110,8 @@ export function doSendDraftTransaction(address, amount) {
       type: ACTIONS.SEND_TRANSACTION_STARTED,
     });
 
-    const successCallback = results => {
-      if (results === true) {
```
seanyesmunt: The response used to just be a boolean

src/redux/reducers/claims.js
```diff
@@ -54,26 +54,17 @@ reducers[ACTIONS.FETCH_CLAIM_LIST_MINE_COMPLETED] = (state, action) => {
   const byId = Object.assign({}, state.byId);
   const pendingById = Object.assign({}, state.pendingById);
 
-  claims
```
seanyesmunt: This is probably the biggest change, now we are only looking at `confirmations` to see if a claim is pending. Before it wouldn't be returned in `claim_list_mine` until it was confirmed so we had to keep track of it ourselves.

src/redux/reducers/claims.js
```diff
@@ -54,26 +54,17 @@ reducers[ACTIONS.FETCH_CLAIM_LIST_MINE_COMPLETED] = (state, action) => {
...
-    .filter(
-      claim => claim.category && (claim.category.match(/claim/) || claim.category.match(/update/))
-    )
-    .forEach(claim => {
+  claims.filter(claim => claim.type && claim.type.match(/claim|update/)).forEach(claim => {
```
skhameneh: Adjust parens:
```claims.filter(claim => claim.type && claim.type.match(/claim|update/))```
to ```claims.filter(claim => ( claim.type && claim.type.match(/claim|update/) ))```

Just for legibility

Comment (skhameneh) [reply src/redux/reducers/claims.js]: You can also combine the filter and forEach to prevent lopping twice

File: src/redux/reducers/claims.js [hunk 5]
```diff
@@ -54,26 +54,17 @@ reducers[ACTIONS.FETCH_CLAIM_LIST_MINE_COMPLETED] = (state, action) => {
...
-      if (pending) {
-        delete pendingById[pending.claim_id];
-      }
-    });
-
-  // Remove old timed out pending publishes
+  // Remove old pending publishes
   Object.values(pendingById)
-    .filter(pendingClaim => Date.now() - pendingClaim.time >= 20 * 60 * 1000)
+    .filter(pendingClaim => byId[pendingClaim.claim_id])
```
skhameneh: Can combine  the filter and forEach to prevent looping twice

src/redux/selectors/claims.js
```diff
@@ -34,8 +33,36 @@ export const selectAllClaimsByChannel = createSelector(
   state => state.claimsByChannel || {}
 );
 
+export const selectPendingById = createSelector(selectState, state => state.pendingById);
+
+export const selectPendingClaims = createSelector(selectState, state =>
```
skhameneh: Would you mind making each arrow fn ` => ` have parens since it's multi-line?
```
export const selectPendingClaims = createSelector(selectState, state => (
  Object.values(state.pendingById || {})
));
```

Not required and this is subjective to preference, let me know what you think, I find it easier to read.

src/redux/selectors/wallet.js
```diff
@@ -130,7 +130,7 @@ export const selectTransactionItems = createSelector(selectTransactionsById, byI
           txid,
           date: tx.timestamp ? new Date(Number(tx.timestamp) * 1000) : null,
           amount,
-          fee: amount < 0 ? (-1 * tx.fee) / append.length : 0,
+          fee: amount < 0 ? -1 * tx.fee / append.length : 0,
```
skhameneh: why remove the parens?

src/redux/selectors/claims.js
seanyesmunt: We can add that, I don't really have a preference. 

I belive it's a prettier option

src/redux/selectors/wallet.js
seanyesmunt: That's just prettier doing it's thing I belive. Although I'm not sure why that is happening now. It should have been run on that file.

PR #86, opened 2018-10-22
add download canceled to the file_info reducer

akinwale:


src/redux/reducers/file_info.js
```diff
@@ -85,6 +85,17 @@ reducers[ACTIONS.DOWNLOADING_PROGRESSED] = (state, action) => {
   });
 };
 
+reducers[ACTIONS.DOWNLOADING_CANCELED] = (state, action) => {
+  const { outpoint } = action.data;
+
+  const newDownloading = Object.assign({}, state.downloadingByOutpoint);
+  delete newDownloading[outpoint];
+
+  return Object.assign({}, state, {
```
skhameneh: Can also use:
```
return {
  ...state,
  {
    downloadingByOutpoint: newDownloading
  }
}
```

PR #87, opened 2018-10-22
modifications for persisting fileList sort

jessopb:


src/redux/actions/file_info.js
```diff
@@ -65,16 +65,9 @@ export function doFetchFileInfosAndPublishedClaims() {
...
-    data: sortOption,
-  };
-}
-
-export function doSetDownloadListSort(sortOption) {
-  return {
-    type: ACTIONS.SET_DOWNLOAD_LIST_SORT,
-    data: sortOption,
+    type: ACTIONS.SET_FILE_LIST_SORT,
+    payload: { page, value },
```
seanyesmunt: This should still be called `data`

src/redux/reducers/file_info.js
```diff
@@ -2,8 +2,9 @@ import * as ACTIONS from 'constants/action_types';
 
 const reducers = {};
 const defaultState = {
-  downloadListSort: '',
-  publishListSort: '',
+  fileListPublishedSort: 'dateNew',
```
seanyesmunt: The options for sorting should be constants

src/redux/reducers/file_info.js
```diff
@@ -158,15 +159,19 @@ reducers[ACTIONS.FETCH_DATE] = (state, action) => {
   return null;
 };
 
-reducers[ACTIONS.SET_PUBLISH_LIST_SORT] = (state, action) =>
-  Object.assign({}, state, {
-    publishListSort: action.data,
-  });
+reducers[ACTIONS.SET_FILE_LIST_SORT] = (state, action) => {
+  const pageSortStates = {
```
seanyesmunt: The pages should be constants too.

seanyesmunt: That are exported so the app can use them

jessopb: Roger. I wouldn't mind centralizing the page -> state dictionary too, but I don't know if it's necessary, or the right place to put it.

src/constants/pages.js
```diff
@@ -0,0 +1,21 @@
+export const AUTH = 'auth';
+export const BACKUP = 'backup';
+export const CHANNEL = 'channel';
+export const DISCOVER = 'discover';
+export const FILE = 'file';
+export const FILE_LIST_DOWNLOADED = 'fileListDownloaded';
+export const FILE_LIST_PUBLISHED = 'fileListPublished';
```
seanyesmunt: This should just be `published`

seanyesmunt:
@jessopb Looks good. I would just make sure we are using the same page values as the app is using. Looks like there are a few differences
https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/component/router/view.jsx#L38-L57

PR #101, opened 2018-11-07
estimate claim date with block height

seanyesmunt:
There are a few formatting changes.

The major change is right here https://github.com/lbryio/lbry-redux/pull/101/files#diff-64ef3f78e6a414fa0a0b21aeabef0db9R217

Also added a small change to fetch transactions after abandoning


src/redux/reducers/wallet.js
```diff
@@ -281,6 +283,11 @@ reducers[ACTIONS.SET_TRANSACTION_LIST_FILTER] = (state: WalletState, action: {})
     transactionListFilter: action.data,
   });
 
+reducers[ACTIONS.UPDATE_CURRENT_HEIGHT] = (state: WalletState, action: { data: number }) =>
+  Object.assign({}, state, {
```
skhameneh: can use parens at the top level for formatting:
```
() => (
  Object.assign({}, {})
);
```

PR #110, opened 2019-01-28
Add new selector for first item in related content

seanyesmunt:
https://github.com/lbryio/lbry-desktop/issues/1948


src/redux/selectors/claims.js
```diff
@@ -4,15 +4,15 @@ import { selectSearchUrisByQuery } from 'redux/selectors/search';
 import { createSelector } from 'reselect';
 import { isClaimNsfw } from 'util/claim';
 
-const selectState = (state) => state.claims || {};
```
seanyesmunt: We added a prettier rule in this repo for this, but since we aren't using it in other projects, i removed it.

PR #115, opened 2019-02-18
fixes for mobile errors

akinwale:


src/redux/selectors/navigation.js
```diff
@@ -5,7 +5,7 @@ export const selectState = state => state.navigation || {};
 
 export const selectCurrentPath = createSelector(selectState, state => state.currentPath);
 
-export const computePageFromPath = path => path.replace(/^\//, '').split('?')[0];
+export const computePageFromPath = path => (path ? path.replace(/^\//, '').split('?')[0] : 1);
```
seanyesmunt: This should probably default to an empty string, since it returns a string when it finds something.

akinwale: Done.

Issue #118, opened 2019-03-03
Remove all api.lbry.io references


tiger5226:
https://github.com/lbryio/lbry-redux/blob/76d8bbef9640bf8ea5c4f45550e55b77d3944ee3/src/lbryapi.js#L9-L11

This repo should be using the `lbryinc` repo to connect to the apis. 

https://github.com/lbryio/lbryinc/blob/f590ddc077bf6e3e8fcc64820ba94a193525d911/src/lbryio.js#L11-L14

A few problems with this:

- Because of environment variables in node modules not working, the following has to be leveraged. The only way to set the url is to modify bundle files. 

https://github.com/lbryio/lbryinc/blob/f590ddc077bf6e3e8fcc64820ba94a193525d911/src/lbryio.js#L11-L14

- Because it is using this old api library it does not send the auth token. I need the `list_homepage` api to receive a token for my discovery changes I am testing. 

As a result this is a blocker for me. I spoke with @seanyesmunt and this was just an oversight that it was not moved before. He is going to move them over asap.

seanyesmunt:
tldr; move homepage/trending to lbryinc

cc @akinwale I'll add them to `lbryinc`, then will create a PR to `lbry-redux` removing them once the android app is moved over.

PR #119, opened 2019-03-10
[BREAKING CHANGE] remove refs to api.lbry.io

seanyesmunt:
There were a few references to api.lbry.io in `lbry-redux`. I've moved these over to `lbryinc` in https://github.com/lbryio/lbryinc/pull/28

This is step one for https://github.com/lbryio/lbry-redux/issues/118


seanyesmunt:
@akinwale I have a build working with these for the desktop. Once they are merged I'll share a commit that makes all the changes. Just a few `import { x } from 'lbry-redux` to `import { x } from 'lbryinc'`

PR #124, opened 2019-03-27
Remove references to desktop app navigation state

seanyesmunt:
Needed for https://github.com/lbryio/lbry-desktop/pull/2384

@akinwale I did a search on `lbry-android`. I don't see any of the functions I've removed, this shouldn't be a breaking change.

### DIFF
--- a/src/redux/actions/search.js
+++ b/src/redux/actions/search.js
@@ -7,7 +7,7 @@ import {
   makeSelectSearchUris,
   selectSuggestions,
   makeSelectQueryWithOptions,
-  selectSearchQuery,
+  selectSearchValue,
 } from 'redux/selectors/search';
 import { batchActions } from 'util/batchActions';
 import debounce from 'util/debounce';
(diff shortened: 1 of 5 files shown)


src/redux/actions/search.js
```diff
@@ -160,15 +157,15 @@ export const doUpdateSearchOptions = (newOptions: SearchOptions) => (
   getState: GetState
 ) => {
   const state = getState();
-  const searchQuery = selectSearchQuery(state);
+  const searchValue = selectSearchValue(state);
```
seanyesmunt: I moved the other functions up because of lint. This is the main change in this file.

src/redux/selectors/cost_info.js
```diff
@@ -0,0 +1,13 @@
+import { createSelector } from 'reselect';
```
seanyesmunt: This file currently exists in `lbryinc`, but should be moved here. I'll keep it there for now

seanyesmunt: I'm dumb. This should continue to exist in `lbryinc`

PR #126, opened 2019-04-03
0.36.0 changes

seanyesmunt:
### Before you look at the diff count...
- Most of these additions are in `flow-typed/npm/` and can be ignored. They come from https://github.com/flow-typed/flow-typed and are auto-generated.
- Most of the other types inside `flow-typed` existed before, but have been checked to match the api in 0.36.0
- The largest actual additions are in `flow-typed/Claim` and `flow-typed/Lbry`

### PR Notes
- Besides the api changes, I went a little crazy with the current Flow setup but even in it's half-finished state, it's so much nicer than what we have currently.
- There are response types for every `Lbry` method.
  - Ideally these would be used automatically but that will be a bit more work.
  - For now you can do `Lbry.status().then((status: StatusResponse) => ...)` and `StatusResponse` will match the response returned from the SDK. Every method with a usable response should have a type that matches the same pattern, ex: `Lbry.claim_list` => `ClaimListResponse`
  - To use them in your project, you need to add a line to your `.flowconfig` like I did here https://github.com/lbryio/lbry-desktop/pull/2407/files#diff-a9d0201e8c5bdc946e54218bb9206174R7

### Breaking changes
- `claim_list_mine` is now `claim_list`
- `claim_tip` is now `support_create`
  - We were already doing this under the hood, but now the method called will actually match the value passed to the sdk
- `doFetchClaimCountByChannel` is removed.
  - The claim count is returned and saved when `resolve` is called with a channel
  - https://github.com/lbryio/lbry-redux/pull/126/files#diff-eaffe64d0c2a6b76072de5084117ddf4R59
- Now only using rollup instead of webpack
  - It was a pain trying to get both to work, rollup is much more preferrable for code splitting.
  - The only changed required to use rollup is: https://github.com/lbryio/lbry-desktop/blob/master/webpack.base.config.js#L98

There are a lot of breaking changes related to the data returned, but those should be the only differences re/ calling methods in `lbry-redux`


.eslintrc.json
```diff
@@ -1,19 +1,14 @@
 {
```
seanyesmunt: We moved the desktop repo over to the same (kinda) rules that spee.ch is using. Keeping this in line with those.

src/redux/selectors/claims.js
```diff
@@ -295,24 +352,25 @@ export const makeSelectRecommendedContentForUri = uri =>
...
+    (claim: ?ChannelClaim, claimsById: { [string]: ChannelClaim | StreamClaim }) => {
+      if (!claim) {
+        return null;
+      }
 
-    return channelName && channelClaimId
-      ? buildURI({ channelName, claimId: channelClaimId }, includePrefix)
-      : null;
-  });
+      // TODO: finish this
```
seanyesmunt: Will finish this when I complete the app changes

seanyesmunt:
@tzarebczan If you want, feel free to check this over too. And leave a comment if you think a certain type will change.

flow-typed/Claim.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+declare type ChannelMetadata = GenericMetadata & {
+  public_key: string,
+  cover_url?: string,
+  contact_email?: string,
+  homepage_url?: string,
+};
+
+declare type StreamMetadata = GenericMetadata & {
```
tzarebczan: Stream will also include the media type data, and have it's own section for the source information: https://github.com/lbryio/types/blob/master/v2/proto/claim.proto#L20

seanyesmunt: Discussed with Tom. This may change, but keeping it for now.

dist/flow-typed/Claim.js
```diff
@@ -0,0 +1,99 @@
...
+  },
+};
+
+declare type Locations =
+  | { latitude: number, longitude: number }
+  | {
+      country?: string,
+      state?: string,
+      code?: string,
+    };
```
akinwale: It makes sense to have `Locations` as a single type with all the fields. `latitude` and `longitude` is applicable to all kinds of locations.

dist/flow-typed/File.js
```diff
@@ -0,0 +1,34 @@
+// @flow
```
akinwale: It looks like the types are being copied to the `dist/` folder. Is this intentional?

src/redux/reducers/claims.js
```diff
@@ -8,26 +9,28 @@ const defaultState = {
 };
 
 reducers[ACTIONS.RESOLVE_URIS_COMPLETED] = (state, action) => {
-  const { resolveInfo } = action.data;
+  const { resolveInfo }: { [string]: ClaimAndCertificate } = action.data;
+  console.log('got em');
```
akinwale: Should be removed before merge, I presume.

src/util/claim.js
```diff
@@ -1,2 +1,25 @@
-export const isClaimNsfw = claim =>
-  claim && claim.value && claim.value.stream && claim.value.stream.metadata.nsfw;
+// @flow
+
+const naughtyTags = ['porn', 'nsfw', 'mature', 'xxx'].reduce(
```
akinwale: Is this always going to be a fixed list? It is advisable to obtain this from an api call because I imagine this could end up being dynamic as the items listed here are by no means exhaustive.

seanyesmunt: That's a good idea. I'll keep it for now just to get things working then do something similar to how we do the blocked outpoint list.

dist/flow-typed/Claim.js
seanyesmunt: Good catch. I'll update.

dist/flow-typed/File.js
seanyesmunt: Yep. Then they can be accessed in apps easily. Ideally the Lbry methods would also have the automatic types added, still working on that.

It makes it so you can add a line to the `flowconfig`, then can just use these as global types.

PR #128, opened 2019-04-08
Fixes bug that caused the app to crash when loading content claims

osilkin98:
This is related to [#2388](https://github.com/lbryio/lbry-desktop/issues/2388) in the `lbry-desktop`, the new daemon broke the ability to load files properly.


seanyesmunt:
Added this in my commit. Closing

PR #133, opened 2019-04-27
set main to bundle.es.js after removing webpack

akinwale:
@seanyesmunt I'll merge this since it's a minor change. Since Webpack is no longer used and a `bundle.js` is no longer generated, I updated `main` in `package.json` to refer to `bundle.es.js` so that React Native would make use of that file instead. If this breaks anything else, please let me know. Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "name": "LBRY Inc.",
     "email": "hello@lbry.io"
   },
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Nice catch. We were doing something else in the desktop repo to handle this but this is a much better solution.

PR #135, opened 2019-05-06
Add cover selector

seanyesmunt:
The same as thumbnail selector, but different. Probably going to add more of these for individual channel metadata


seanyesmunt:
Merging now since it's the exact same as the thumbnail selector

PR #141, opened 2019-05-08
Add comic book media-type detection

btzr-io:
> A comic book archive or comic book reader file (also called sequential image file) is a type of archive file for the purpose of sequential viewing of images, commonly for comic books.
https://en.wikipedia.org/wiki/Comic_book_archive

Required for https://github.com/lbryio/lbry-desktop/issues/1377


btzr-io:
Never mind looks like the desktop app uses this function instead: 
https://github.com/lbryio/lbry-desktop/blob/master/src/ui/util/get-media-type.js

PR #144, opened 2019-05-10
store user supports

seanyesmunt:
The new sdk removed supports from `claim_list`, so we have to fetch and store them on our own


src/redux/reducers/wallet.js
```diff
@@ -64,226 +67,250 @@ const defaultState = {
...
+      });
+
+      return {
+        ...state,
+        transactions: byId,
+        fetchingTransactions: false,
+      };
+    },
+
+    [ACTIONS.FETCH_SUPPORTS_STARTED]: (state: WalletState) => ({
```
seanyesmunt: I just moved the old reducer functions to use the `handleActions` util. The actual changes I made are from here.
L92 - L133

seanyesmunt:
@akinwale not sure you can unlock tips in the android app, but if you can, you will need to use this.

Desktop commit to work with it: https://github.com/lbryio/lbry-desktop/commit/6f1201d898caa53b50e0be340b945761418a144b

src/redux/selectors/wallet.js
```diff
@@ -82,7 +82,12 @@ export const selectTotalBalance = createSelector(
 
 export const selectTransactionsById = createSelector(
   selectState,
-  state => state.transactions
+  state => state.transactions || {}
+);
+
+export const selectSupportsById = createSelector(
+  selectState,
+  state => state.supports || {}
```
kauffj: If we set the default state values to `{}`, are the or clauses lines really necessary? (more curiosity than an objection)

seanyesmunt: I've been doing that to safeguard if the value ends up being set to `null` or `undefined` for some reason. Not that I noticed it happening, but just in case.

PR #147, opened 2019-05-15
Add tags to redux

seanyesmunt:
Will need to make another PR that addresses pagination on the trending pages.


src/index.js
```diff
@@ -127,6 +136,7 @@ export {
   makeSelectCoverForUri,
   makeSelectTitleForUri,
   makeSelectDateForUri,
+  makeSelectTagsForUri,
```
seanyesmunt: Not sure if this should be here or in the tags file. Currently, all metadata selectors are inside of the claims file.

src/constants/tags.js
```diff
@@ -0,0 +1,21 @@
...
+  'economics',
+  'sports',
+  'food',
+  'science',
+  'art',
+  'nature',
+  'beliefs',
+  'music',
+  'pop culture',
+  'weapons',
```
seanyesmunt: Will this be the list of default tags? Either way, just using this for more accurate test data.

flow-typed/Tags.js
```diff
@@ -0,0 +1,30 @@
+declare type TagState = {
+  followedTags: FollowedTags,
+  knownTags: KnownTags,
+  trending: Array<string>,
```
kauffj: if I'm reading this code correctly, this essentially to store the uris for the last call to claim_search

if that's the case, should this state even be in Tags.js instead of Claims.js? and either way, a name like lastClaimSearchUris would probably be better

flow-typed/Tags.js
```diff
@@ -0,0 +1,30 @@
...
+  fetchingTrending: boolean,
```
kauffj: Similar to above, this value is tracking whether there is a claim_search active, which isn't necessarily trending (it could be best, or newest, etc.).

flow-typed/Tags.js
```diff
@@ -0,0 +1,30 @@
...
+declare type FollowedTags = Array<string>;
+
+declare type TagAction = {
+  type: string,
+  data: {
+    name: string,
+  },
+};
+
+declare type TrendingTagAction = {
```
kauffj: Same comment as above, not necessarily "Trending"

src/constants/tags.js
kauffj: The default tags will be close to this, but could they be put in the app codebase rather than lbry-redux?

src/index.js
kauffj: I think they should continue to be in the claims file.

src/redux/actions/tags.js
```diff
@@ -0,0 +1,51 @@
...
+});
+
+export const doDeleteTag = (name: string) => ({
+  type: ACTIONS.TAG_DELETE,
+  data: {
+    name,
+  },
+});
+
+export const doFetchByTags = (amount: number = 10, options: Object = {}) => {
```
kauffj: this looks like it just calls right through to claim search... do we even need this along with the accompanying actions? could everything just call `doClaimSearch` directly?

src/redux/selectors/file_info.js
```diff
@@ -228,3 +228,9 @@ export const selectFileListDownloadedSort = createSelector(
   selectState,
   state => state.fileListDownloadedSort
 );
+
+export const selectDownloadedUris = createSelector(
+  selectFileInfosDownloaded,
+  // We should use permament_url but it doesn't exist in file_list
```
kauffj: should this be filed?

src/constants/tags.js
seanyesmunt: Yes

src/redux/selectors/file_info.js
seanyesmunt: It is.

flow-typed/Tags.js
seanyesmunt: Yeah I'll move it to the claims reducer

seanyesmunt:
Merging since so that I can rebase the comments PR. I responded/updated code based on the comments above. I can make another PR if there are more comments.

PR #148, opened 2019-05-16
redux comments DRAFT WIP

jessopb:
Adds comment functionality to redux


src/redux/actions/comments.js
```diff
@@ -0,0 +1,82 @@
...
+    const claim = selectClaimsByUri(state)[uri];
+    const claimId = claim ? claim.claim_id : null;
+
+    dispatch({
+      type: ACTIONS.COMMENT_LIST_STARTED,
+    });
+    Lbry.comment_list({
+      claim_id: claimId,
+    })
+      .then(results => {
```
seanyesmunt: This is a `CommentListResponst` type

src/redux/actions/comments.js
```diff
@@ -0,0 +1,82 @@
...
+        dispatch({
+          type: ACTIONS.COMMENT_LIST_COMPLETED,
+          data: {
+            comments: results,
+            claimId: claimId,
+            uri: uri,
+          },
+        });
+      })
+      .catch(error => {
```
seanyesmunt: We should have a `COMMENT_LIST_FAILED` action that handles this error. (we don't need to display the error in the ui right now)

src/redux/actions/comments.js
```diff
@@ -0,0 +1,82 @@
...
+      channel_id,
+    })
+      .then((result: Comment) => {
+        dispatch({
+          type: ACTIONS.COMMENT_CREATE_COMPLETED,
+          data: {
+            response: result,
+          },
+        });
+        dispatch({
```
seanyesmunt: Whatever happens in the `COMMENT_LIST_UPDATED` handler can just be done in `COMMENT_CREATE_COMPLETED`.

Just append the new comment there, then we don't need two actions.

src/redux/reducers/comments.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+import * as ACTIONS from 'constants/action_types';
+
+// TODO change to handleActions()
```
seanyesmunt: 🕺

src/redux/reducers/comments.js
```diff
@@ -0,0 +1,66 @@
...
+  isLoading: false,
+};
+
+reducers[ACTIONS.COMMENT_CREATE_STARTED] = (state: CommentsState, action: any): CommentsState =>
+  Object.assign({}, state, {
+    isLoading: true,
+  });
+
+reducers[ACTIONS.COMMENT_CREATE_FAILED] = (state: CommentsState, action: any): CommentsState => {
+  // TODO: handle error.. what else?
```
seanyesmunt: is `isLoading` to `false`

src/redux/reducers/comments.js
```diff
@@ -0,0 +1,66 @@
...
+  return state;
+};
+
+reducers[ACTIONS.COMMENT_CREATE_COMPLETED] = (state: CommentsState, action: any): CommentsState => {
+  const { comments }: any = action.data;
```
seanyesmunt: Add this new comment to the list of comments here.

dist/bundle.es.js
```diff
@@ -2769,6 +2778,74 @@ const doDeleteTag = name => ({
...
+}
+
+function doCommentCreate(comment = '', claim_id = '', channel, parent_id) {
+  return (dispatch, getState) => {
+    const state = getState();
+    dispatch({
+      type: COMMENT_CREATE_STARTED
+    });
+    const myChannels = selectMyChannelClaims(state);
+    const namedChannelClaim = myChannels.find(myChannel => myChannel.name === channel);
```
osilkin98: The SDK already does this anyway when signing the comments

PR #150, opened 2019-05-21
fix:  claim search

tzarebczan:
to use channel uri instead of claim id (support for vanity resolution)

I did not test this locally - would need to bring it into an Android build. The --winning parameter is required so that only the winning channel's claims are returned.


tzarebczan:
This will be changing to --channel and --is_controlling in 0.38, but this is correct for 0.37 which the Android app should ship with @akinwale

PR #152, opened 2019-05-31
adds headers and means to set them to Lbry apicall function

jessopb:


dist/bundle.es.js
```diff
@@ -606,12 +606,16 @@ const Lbry = {
   isConnected: false,
   connectPromise: null,
   daemonConnectionString: 'http://localhost:5279',
+  apiHeaders: { 'content-type': 'application/json-rpc' },
```
kauffj: mildly in favor of calling this `apiRequestHeaders`

src/lbry.js
```diff
@@ -11,12 +11,16 @@ const Lbry: LbryTypes = {
...
   connectPromise: null,
   daemonConnectionString: 'http://localhost:5279',
+  apiHeaders: { 'content-type': 'application/json-rpc' },
 
   // Allow overriding daemon connection string (e.g. to `/api/proxy` for lbryweb)
   setDaemonConnectionString: (value: string) => {
     Lbry.daemonConnectionString = value;
   },
 
+  addApiHeaders: (headersObj: {}) => {
```
kauffj: There should presumably be a remove method as well.

I'm ambivalent between using `add` (add `n`) vs `set` (set 1).

kauffj:
There's no need to submit this for a second review, just merge the changes.

PR #153, opened 2019-06-04
updates for sdk@38

seanyesmunt:


src/redux/actions/claims.js
```diff
@@ -34,29 +34,42 @@ export function doResolveUris(uris: Array<string>, returnCachedClaims: boolean =
...
-        } else {
-          // $FlowFixMe
-          const { claim, certificate, claims_in_channel: claimsInChannel } = uriResolveInfo;
-          resolveInfo[uri] = { claim, certificate, claimsInChannel };
+        if (uriResolveInfo) {
+          if (uriResolveInfo.error) {
+            resolveInfo[uri] = { ...fallbackResolveInfo };
+          } else {
+            let result = {};
+            if (uriResolveInfo.value_type === 'channel' ) {
```
kauffj: This code seems fine from the app side. Is there any feedback for the SDK team on these signatures?

src/redux/reducers/claims.js
```diff
@@ -46,36 +46,42 @@ const defaultState = {
...
-      byUri[uri] = claim.claim_id;
-
-      byId[certificate.claim_id] = certificate;
-      const channelUri = `lbry://${certificate.name}#${certificate.claim_id}`;
-      byUri[channelUri] = certificate.claim_id;
-    } else if (!claim && certificate) {
-      byId[certificate.claim_id] = certificate;
-      byUri[uri] = certificate.claim_id;
+  Object.entries(resolveInfo).forEach(([uri, { channel, stream }]) => {
+    if (stream && !channel) {
```
kauffj: I think below is equivalent but simpler and less repetitive: 

```
if (stream) {
      byId[stream.claim_id] = stream;
      byUri[uri] = stream.claim_id;
}
if (channel) {
      byId[channel.claim_id] = channel;
      byUri[stream ? channel.permanent_url : uri] = channel.claim_id;
}
if (!stream && !channel) {
      byUri[uri] = null;
}
```

src/redux/actions/claims.js
seanyesmunt: A little bit. The responses are becoming more similar between different sdk methods, which helps a lot.

PR #157, opened 2019-06-28
Discovery fixes

seanyesmunt:


dist/bundle.es.js
```diff
@@ -872,32 +872,37 @@ function parseURI(URI, requireProto = false) {
 
   // Validate protocol
   if (requireProto && !proto) {
-    throw new Error(__('LBRY URIs must include a protocol prefix (lbry://).'));
+    console.error(__('LBRY URIs must include a protocol prefix (lbry://).'));
```
kauffj: Why is this better? Won't anything using this choke on receiving an empty object anyway?

src/redux/selectors/claims.js
```diff
@@ -256,12 +256,16 @@ export const selectMyClaims = createSelector(
...
+  myClaims =>
+    myClaims.filter(claim => !claim.name.match(/^@/)).sort((a, b) => a.timestamp - b.timestamp)
 );
 
 export const selectMyClaimUrisWithoutChannels = createSelector(
   selectMyClaimsWithoutChannels,
-  myClaims => myClaims.map(claim => `lbry://${claim.name}#${claim.claim_id}`)
+  myClaims =>
+    myClaims
+      .sort((a, b) => a.confirmations - b.confirmations)
```
kauffj: It's a corner case, but there's nothing stopping two publishes from being in the same block.

kauffj: Particularly with YT sync publishes, this is likely.

src/redux/selectors/file_info.js
```diff
@@ -232,5 +232,9 @@ export const selectFileListDownloadedSort = createSelector(
 export const selectDownloadedUris = createSelector(
   selectFileInfosDownloaded,
   // We should use permament_url but it doesn't exist in file_list
-  info => info.map(claim => `lbry://${claim.claim_name}#${claim.claim_id}`)
+  info =>
+    info
+      .slice()
+      .reverse()
+      .map(claim => `lbry://${claim.claim_name}#${claim.claim_id}`)
```
kauffj: Is the comment here still true, that the URL isn't available on the claim?

seanyesmunt: Yep

dist/bundle.es.js
seanyesmunt: I guess they could, probably better to just fix it in the app. I'll remove

src/redux/selectors/claims.js
seanyesmunt: Yep, I should have been using `timestamp`

seanyesmunt:
@kauffj Going to merge this so I can merge the `lbry-desktop` pr (aka I already merged that because I forgot this PR needed to be merged too)

Let me know if you have more comments.

PR #158, opened 2019-06-28
Publishing

akinwale:


kauffj:
@akinwale @jessopb would like this merged, any reason it cannot be?

akinwale:
No reason. I'll merge it now.

src/redux/actions/publish.js
```diff
@@ -0,0 +1,404 @@
...
+      publishData.filePath = fileInfo.download_path;
+    } catch (e) {
+      console.error(e.name, e.message);
+    }
+  }
+
+  dispatch({ type: ACTIONS.DO_PREPARE_EDIT, data: publishData });
+};
+
+export const doPublish = (params: PublishParams) => (dispatch: Dispatch, getState: () => {}) => {
```
seanyesmunt: Can this be updated to use the new logic in `doPublish` for the desktop app?

It removes the need to pass all of the values to `doPublish` since we already store them in state.

https://github.com/lbryio/lbry-desktop/blob/master/src/ui/redux/actions/publish.js#L208-L250

@jessopb maybe you could do this?

jessopb: So we expect every actionListener to call doUpdatePublishForm and I add the selector into the doPublish action.
I also had to add makeSelectPublishFormValue item => ... for desktop not to crash.

PR #165, opened 2019-07-15
support multiple simultaneous claim_search requests

akinwale:


seanyesmunt:
@akinwale I'm going to close my PR and use this. I'll make the changes needed to work with different sorting as well.

PR #166, opened 2019-07-17
cache claim search results by query

seanyesmunt:
#### Changes
- Change claim_search cache to use the exact options that were used for the claim search to create a key, instead of just tags
  - This requires directly passing in the exact `options` into `claimSearch` you want to search with

- also a couple bugs that we fixed before the last desktop release


seanyesmunt:
I'm going to bring these changes into #165 since we don't need both

src/redux/selectors/claims.js
```diff
@@ -88,12 +88,14 @@ export const makeSelectClaimForUri = (uri: string) =>
...
 
+      let valid;
       let claimId;
       try {
         ({ claimId } = parseURI(uri));
+        valid = true;
       } catch (e) {}
 
-      if (claimId) {
+      if (valid) {
```
kauffj: I'm mildly confused by why this is necessary, but not a blocking comment.

src/redux/reducers/claims.js
```diff
@@ -44,8 +48,9 @@ const defaultState = {
   fetchingMyChannels: false,
   abandoningById: {},
   pendingById: {},
+  claimSearchError: false,
   fetchingClaimSearch: false,
-  lastClaimSearchUris: [],
+  claimSearchSearchByQuery: {},
```
kauffj: what triggers invalidation of this cache? what is the behavior if there are weeks or months old results in here?

src/redux/actions/publish.js
```diff
@@ -375,7 +385,9 @@ export const doCheckPendingPublishes = () => (dispatch: Dispatch, getState: GetS
           // If it's confirmed, check if we should notify the user
           if (selectosNotificationsEnabled(getState())) {
             const notif = new window.Notification('LBRY Publish Complete', {
-              body: `${claim.value.title} has been published to lbry://${claim.name}. Click here to view it`,
+              body: `${claim.value.title} has been published to lbry://${
```
kauffj: Missing `__`

kauffj:
Discussions on standup have led me to believe this needs to be thought out more.

src/redux/selectors/claims.js
seanyesmunt: It fixed... something. I believe selecting vanity urls

src/redux/selectors/navigation.js
```diff
@@ -1,50 +1,78 @@
...
 export const selectActiveHistoryEntry = createSelector(
   selectState,
   state => state.stack[state.index]
 );
 
-export const selectPageTitle = createSelector(selectCurrentPage, page => {
-  switch (page) {
-    default:
-      return '';
+export const selectPageTitle = createSelector(
```
kauffj: is this actually used?

src/redux/selectors/claims.js
```diff
@@ -253,7 +255,9 @@ export const makeSelectThumbnailForUri = (uri: string) =>
     makeSelectClaimForUri(uri),
     claim => {
       const thumbnail = claim && claim.value && claim.value.thumbnail;
-      return thumbnail && thumbnail.url && thumbnail.url.trim().length > 0 ? thumbnail.url : undefined;
+      return thumbnail && thumbnail.url && thumbnail.url.trim().length > 0
+        ? thumbnail.url
+        : undefined;
```
kauffj: this means that if `thumbnail.url` is `''` or `null`, we return `undefined`

this seems wrong

`undefined` seems okay if `thumbnail.url` is undefined itself (i.e. `thumbnail` is null, not an object, or does not have the `url` property set)

src/redux/selectors/navigation.js
seanyesmunt: No. And I think others here can be removed

src/redux/selectors/claims.js
seanyesmunt: So just return `""` if it's an empty string? Instead of `undefined`?

src/redux/selectors/navigation.js
seanyesmunt: Yeah, we weren't even exporting this file. Just deleted it.

PR #172, opened 2019-07-24
supports channel page

jessopb:
adds selectMyChannelUris to claim reducer


jessopb:
I'll just merge this into cache branch instead.

PR #173, opened 2019-07-30
Channel filter 

jessopb:
Supports redux so that users can block channels and visit their blocked channels page through settings.


jessopb:
Closed for #175

PR #175, opened 2019-08-02
Channel filter mm

jessopb:
provides user capability to block channels

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -273,4 +273,4 @@ export {
 
 export { selectFollowedTags, selectUnfollowedTags } from 'redux/selectors/tags';
 
-export { selectBlockedChannels, selectChannelIsBlocked } from 'redux/selectors/blocked';
+export { selectBlockedChannels, selectChannelIsBlocked, selectBlockedChannelsCount } from 'redux/selectors/blocked';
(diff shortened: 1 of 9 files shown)


src/index.js
```diff
@@ -127,6 +129,7 @@ export { notificationsReducer } from 'redux/reducers/notifications';
 export { publishReducer } from 'redux/reducers/publish';
 export { searchReducer } from 'redux/reducers/search';
 export { tagsReducer } from 'redux/reducers/tags';
+export { blockChannelReducer } from 'redux/reducers/blocked';
```
seanyesmunt: This should probably just be `blockedReducer` since we will most likely add blocked tags to this too.

src/redux/selectors/blocked.js
```diff
@@ -0,0 +1,19 @@
...
+);
+
+export const selectBlockedChannelsCount = createSelector(
+  selectState,
+  (state: BlocklistState) => state.blockedChannels.length
+);
+
+export const selectChannelIsBlocked = (uri: string) => createSelector(
+  selectState,
+  (state: BlocklistState) => { return state.blockedChannels.includes(uri) }
```
seanyesmunt: Not a big deal but this should probably use `selectBlockedChannels` instead of `selectState`

PR #176, opened 2019-08-02
Range request support for desktop

seanyesmunt:
#### Changes
- Update `fileInfo` after successful `get` to populate with `streaming_url`
- Removed `purchasedStreamingUrls` from redux, this can now just be accessed from `fileInfo`

#### Notes
@akinwale I still think it makes sense to combine `file` and `file_info` reducers. I think it could be valuable to keep purchase history separate, so we can persist that without persisting a user's entire `file_list` (maybe?). Maybe `file` should be renamed to `purchase`? Or something similar. Not a big deal now though.


src/redux/selectors/file.js
```diff
@@ -31,8 +32,10 @@ export const selectLastPurchasedUri: (state: State) => string = createSelector(
...
 );
 
-export const makeSelectStreamingUrlForUri = (uri: string): ((state: State) => {}) =>
+export const makeSelectStreamingUrlForUri = (uri: string) =>
   createSelector(
-    selectPurchasedStreamingUrls,
-    streamingUrls => streamingUrls && streamingUrls[uri]
+    makeSelectFileInfoForUri(uri),
+    fileInfo => {
+      return fileInfo && fileInfo.streaming_url;
```
seanyesmunt: @akinwale Do we even need to keep track of `streamingUrls` if we can just access it in `fileInfo`?

src/lbry.js
```diff
@@ -32,27 +33,40 @@ const Lbry: LbryTypes = {
...
   // Returns a human readable media type based on the content type or extension of a file that is returned by the sdk
-  getMediaType: (contentType: string, extname: ?string) => {
-    if (extname) {
-      const formats = [
-        [/^(mp4|m4v|webm|flv|f4v|ogv)$/i, 'video'],
-        [/^(mp3|m4a|aac|wav|flac|ogg|opus)$/i, 'audio'],
-        [/^(html|htm|xml|pdf|odf|doc|docx|md|markdown|txt|epub|org)$/i, 'document'],
-        [/^(stl|obj|fbx|gcode)$/i, '3D-file'],
-      ];
+  getMediaType: (contentType?: string, fileName: ?string) => {
```
seanyesmunt: I guess there have been some changes to this in the desktop codebase and we never moved over. Just copy pasting this in so we can switch to it.

Android doesn't use the second argument for this function so changing it should be fine.

tzarebczan: I also mentioned to you that the SDK now returns a stream type: 
```  ...
      },
      "stream_type": "video",
      ],
```

The mapping is here - I think it should be able to work with some small adjustments (glaringly missing are fbx/gcode, but not sure if anyone uses those): https://github.com/lbryio/lbry-sdk/blob/41b866c1eeb002aaa0e42bc71734f6879d3b4457/lbry/lbry/schema/mime_types.py

kauffj:
@shockr can you review/merge before 11am meeting tomorrow?

src/redux/actions/file.js
```diff
@@ -9,10 +9,10 @@ import { makeSelectStreamingUrlForUri } from 'redux/selectors/file';
 type Dispatch = (action: any) => any;
 type GetState = () => { file: FileState };
 
-export function doFileGet(uri: string, saveFile: boolean = true) {
+export function doFileGet(uri: string, saveFile: boolean = true, onSuccess?: GetResponse => any) {
   return (dispatch: Dispatch) => {
     dispatch({
-      type: ACTIONS.LOADING_FILE_STARTED,
+      type: ACTIONS.PURCHASE_URI_STARTED,
```
akinwale: Since `doPurchaseUri` also dispatches the same action, how do we differentiate between the two? I recall using a different action here because it's possible for a purchase to fail due to insufficient credits, the wallet being locked or some other condition, so `get` doesn't actually get called in those scenarios.

src/redux/selectors/file_info.js
```diff
@@ -72,7 +74,7 @@ export const selectFileInfosDownloaded = createSelector(
       return (
         fileInfo &&
         myClaimIds.indexOf(fileInfo.claim_id) === -1 &&
-        (fileInfo.completed || fileInfo.written_bytes)
+        (fileInfo.completed || fileInfo.written_bytes > 0 || fileInfo.blobs_completed > 0)
```
akinwale: Can a file be considered downloaded where just `blobs_completed > 0`? I believe it's also possible for streamed files to meet the condition `blobs_completed > 0`, but I'll have to test this to be sure.

seanyesmunt: The way we have been using this selector in the past is `is any part of it downloaded`. Possibly we should add another called `selectFileInfosCompletelyDownloaded`? Or something similar.

src/redux/actions/file.js
seanyesmunt: Ah that's a good point. I just searched for `LOADING_FILE_` and didn't find anything so I figured it wasn't needed. That does make sense to have two though. I'll add it back. Maybe `GET_FILE_XXX` to signal that it's specifically for `get` calls?

akinwale: `GET_FILE_XXX` sounds perfect.

src/redux/selectors/file_info.js
akinwale: That makes sense so it's fine as is. For completely downloaded, the condition would have to be something like `fileInfo.completed || fileInfo.written_bytes >= fileInfo.total_bytes`.

PR #178, opened 2019-08-11
add selectors for total supports and claim supports

zxawry:
I've added two selectors, one for sum of all active supports on all claims (`selectTotalSupports`) I think can be used to show total revocable supports on wallet page as well as balance.
and another for sum of supports on a specific claim (`makeSelectSupportsForUri`) to show tips and supports added to an own claim.


src/redux/selectors/claims.js
```diff
@@ -531,3 +532,24 @@ export const makeSelectShortUrlForUri = (uri: string) =>
...
+    makeSelectClaimForUri(uri),
+    (byOutpoint, claim: ?StreamClaim) => {
+      if (!claim || !claim.is_mine) {
+        return null;
+      }
+
+      const { claim_id: claimId } = claim;
+      let total = parseFloat("0.0");
+
+      Object.values(byOutpoint).forEach(support => {
```
seanyesmunt: This can just use `reduce`

seanyesmunt:
This is great @zxawry!

Can you run `yarn build` so the bundle is updated with this?

zxawry:
Done! I thought it may cause conflicts.

PR #179, opened 2019-08-13
Mobile updates

akinwale:


akinwale:
@jessopb @seanyesmunt This removes the `fs` and `path` imports as they're not available on mobile. They will need to be passed in as arguments to the functions calling them.

PR #184, opened 2019-08-15
improve short format for balance credits >1K and >1M

akinwale:


src/util/format-credits.js
```diff
@@ -5,6 +5,21 @@ export function formatCredits(amount, precision) {
     .replace(/\.?0+$/, '');
 }
 
+export function formatBigNumberCredits(amount, precision) {
```
kauffj: Would it make more sense for this functionality to be a parameter to formatCredits rather than a standalone function?

akinwale: I was trying to avoid having to update all the references to the method, as there are quite a few.

kauffj: Why would you have to update the references if you add an optional 3rd parameter?

akinwale: I suppose I wouldn't have to, now that I think about it. I should be able to roll the functionality into the `formatCredits` method.

PR #187, opened 2019-08-19
use canonical_url for everything

seanyesmunt:
This populates `byUri` with `claim.canonical_url` instead of `claim.permanent_url` everywhere.

Once both apps are moved over, the search results should move to canonical_url too:
https://github.com/lbryio/lighthouse/issues/163

### lbryURI.js
- The previous version of `buildURI` and `parseURI` did not work properly with canonical_urls
- `buildURI` and `parseURI` will continue to work the same (which will be unreliable if using claim.canonical_url), but will blow up your console with errors in dev mode if `claimId`, `claimName` or `contentName` is passed in. These should be replaced with `{stream}/{channel}Name`, `{stream}/{channel}ClaimId`.


src/lbryURI.js
```diff
@@ -64,10 +70,13 @@ export function parseURI(URI, requireProto = false) {
       throw new Error(__(`Channel names must be at least %s characters.`, channelNameMinLength));
     }
 
-    contentName = path;
+    // Probably need to fix the components regex to handle claim ids attached to contentNames when channels are included
```
seanyesmunt: Not sure the best way to do this.

Should `path` include the claimId? Should there be `path` and `pathClaimid`?

src/lbryURI.js
```diff
@@ -64,10 +65,13 @@ export function parseURI(URI, requireProto = false) {
...
+    if (path) {
+      contentName = path.replace('#', '');
+    }
   }
 
-  const nameBadChars = (channelName || claimName).match(regexInvalidURI);
+  const nameBadChars = (contentName || isChannel ? channelName : claimName).match(regexInvalidURI);
```
kauffj: This could possibly be removed or moved. The function is called parseURI, not parseAndValidateURI. It's one thing to error if invalid characters make it impossible to parse, but I think erroring on a validly structured URI is probably incorrect here.

src/lbryURI.js
kauffj: According to https://lbry.tech/spec#urls, the `path` should include the channel name, claim name, and any modifiers. The path is everything between the scheme and any query parameters.

Channel name, claim name, and any modifiers would be sub-properties of the path (and could potentially be returned/parsed out by this function as well).

@lyoshenka, please correct me if I'm wrong about this.

@seanyesmunt we should try to have this function use the same names and standards that the spec specifies. If you think we should do something different than what the spec says, we can also consider amending the spec.

src/redux/actions/claims.js
```diff
@@ -329,9 +329,12 @@ export function doClaimSearch(
     const success = (data: ClaimSearchResponse) => {
       const resolveInfo = {};
       const uris = [];
+      const canonicalUris = [];
```
kauffj: why keep two sets of URIs?

src/redux/reducers/claims.js
```diff
@@ -66,24 +68,31 @@ function handleClaimAction(state: State, action: any): State {
...
       byUri[uri] = stream.claim_id;
+
+      // byUri[uri] should be removed in favor of this as apps move to use canonical_url everywhere
+      // @canonicalfixme
+      byUri[stream.canonical_url] = stream.claim_id;
     }
+
     if (channel) {
       byId[channel.claim_id] = channel;
+      byUri[stream ? channel.canonical_url : uri] = channel.claim_id;
```
kauffj: URI or URL? are we consistent with when we use one vs. the other?

src/redux/selectors/claims.js
```diff
@@ -511,31 +511,32 @@ export const selectFetchingClaimSearch = createSelector(
   fetchingClaimSearchByQuery => Boolean(Object.keys(fetchingClaimSearchByQuery).length)
 );
 
+// @canonicalfixme
 export const selectClaimSearchByQuery = createSelector(
   selectState,
   state => state.claimSearchByQuery || {}
 );
 
+export const selectClaimSearchByQueryCanonical = createSelector(
```
kauffj: IMO, all of @lbryio/appsquad should be on the same page with the URI type that is promised/implied by default when used without modifiers. Is it the user input URI, the canonical URI, the permanent URI (canonical means something close to permanent so I dislike using both of these as adjectives), or something else?

Then when we have structures like `xxxByUri`, we understand that it is implicitly promising an index by that URI type, without needing to always specify which type, and specifying only when we deviate from the default.

src/lbryURI.js
seanyesmunt: I like removing validation from this. We already have `isURIValid` to validate

seanyesmunt:
@kauffj this is ready for another review. 

It's a bigger change than I wanted to make but it's already making some of the url code way easier to follow without trying to figure out if a name is for the channel or stream. Let me know what you think.

src/redux/actions/search.js
```diff
@@ -112,18 +112,26 @@ export const doSearch = (
...
-        if (result.name) {
-          const uri = buildURI({
-            claimName: result.name,
-            claimId: result.claimId,
-          });
-          actions.push(doResolveUri(uri));
-          uris.push(uri);
+        if (result) {
+          const { name, claimId } = result;
+          const urlObj = {};
```
kauffj: should this be a Flow type?

PR #192, opened 2019-08-29
fix: canonical URL stuff

tzarebczan:
Added cross-env to support compiling on Windows
Channel name now returns the channel with @
updated remaining instances of claimid/contentname
removed unused function that relied on parseURI


tzarebczan:
Fixed

PR #193, opened 2019-08-30
canonical_url fixes

seanyesmunt:
Random fixes for canonical_url


seanyesmunt:
Merging this since it's nothing major and the desktop app was released with it.

PR #195, opened 2019-09-02
add optionalParams to doCreateChannel. Update myClaims after fetching channels

akinwale:


dist/bundle.es.js
```diff
@@ -2394,16 +2398,40 @@ function doFetchClaimsByChannel(uri, page = 1) {
...
+    const createParams = {
       name,
       bid: creditsToString(amount)
-    })
+    };
+
+    if (optionalParams.title) {
+      createParams.title = optionalParams.title;
+    }
+    if (optionalParams.cover) {
```
seanyesmunt: Should this be called `coverUrl` to match the sdk value?

akinwale: I suppose it should. I'll also have to update this in `doUpdateChannel`.

PR #196, opened 2019-09-05
restores image viewer

jessopb:
Viewing Brendan's gnarly sunset, clicking the eyeball told me the viewer couldn't view it. Are there other formats that should be added for images, or is there another way that's better to handle images?

### DIFF
--- a/src/lbry.js
+++ b/src/lbry.js
@@ -36,6 +36,7 @@ const Lbry: LbryTypes = {
     if (fileName) {
       const formats = [
         [/\.(mp4|m4v|webm|flv|f4v|ogv)$/i, 'video'],
+        [/\.(jpg|jpeg|png|gif)$/i, 'image'],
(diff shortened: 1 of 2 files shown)


src/lbry.js
```diff
@@ -36,6 +36,7 @@ const Lbry: LbryTypes = {
     if (fileName) {
       const formats = [
         [/\.(mp4|m4v|webm|flv|f4v|ogv)$/i, 'video'],
+        [/\.(jpg|jpeg|png|gif)$/i, 'image'],
         [/\.(mp3|m4a|aac|wav|flac|ogg|opus)$/i, 'audio'],
         [/\.(jpeg|jpg|png|gif|svg)$/i, 'image'],
```
seanyesmunt: Might need to rebase...

:)

seanyesmunt:
This is already fixed.

src/lbry.js
jessopb: I fixed it, then rebased it :/

PR #197, opened 2019-09-09
Onboarding updates

seanyesmunt:
#### Changes
- `myChannelClaims` now defaults to `undefined`. This is helpful to know if we've fetched a users channel claims or not. `undefined` === not fetched, `null` === fetched and no channels
- I added a `sync` file here. I'm not sure if it belongs here but I think it makes sense to handle the data (after it's been fetched by some service) to populate our reducers in `lbry-redux`. If everything is in `lbryinc` we need to either import a constant from `lbryinc` into `lbry-redux` or just hard-code the `ACTIONS.POPULATE_USER_SETTINGS` action.


src/redux/actions/sync.js
```diff
@@ -0,0 +1,29 @@
...
+
+type v0Data = {
+  version: '0',
+  shared: {
+    subscriptions?: Array<string>,
+    tags?: Array<string>,
+  },
+};
+
+function extractSettings(rawObj: v0Data) {
```
kauffj: replace settings with shared user state, shared state, or something else

src/redux/reducers/claims.js
```diff
@@ -42,14 +44,16 @@ const defaultState = {
   resolvingUris: [],
   // This should not be a Set
   // Storing sets in reducers can cause issues
-  myChannelClaims: new Set(),
+  myChannelClaims: undefined,
```
kauffj: why undefined though?

kauffj: if Sets have issues, just make it an object and use the keys, it's basically the same thing

src/redux/reducers/tags.js
```diff
@@ -65,6 +65,31 @@ export const tagsReducer = handleActions(
         followedTags: newFollowedTags,
       };
     },
+    [ACTIONS.USER_STATE_POPULATE]: (
```
kauffj: when receiving saved/stored user state, I think it's probably fine (and possible better) to overwrite than to try to merge

...I may be misunderstanding what this function is doing

src/redux/selectors/claims.js
```diff
@@ -350,9 +360,12 @@ export const selectMyChannelClaims = createSelector(
   selectState,
   selectClaimsById,
   (state, byId) => {
-    const ids = state.myChannelClaims || [];
-    const claims = [];
+    const ids = state.myChannelClaims;
+    if (!ids) {
+      return ids;
```
kauffj: this should probably be `return []`;

src/redux/reducers/claims.js
seanyesmunt: So we can track if we've fetched a users channels. If it's undefined, we haven't fetched them yet.

src/redux/selectors/claims.js
seanyesmunt: It's returning undefined so we can check in the app if we've ever fetched before.

akinwale:
@seanyesmunt I merged the latest commit with the master branch in order to get the changes that I merged from the `channel-creator` branch in. If this causes any problems, please feel free to revert. Thanks.

src/redux/reducers/tags.js
```diff
@@ -65,6 +65,16 @@ export const tagsReducer = handleActions(
...
+      state: TagState,
+      action: { data: { tags: ?Array<string> } }
+    ) => {
+      const { tags } = action.data;
+      return {
+        ...state,
+        followedTags: tags && tags.length ? tags : DEFAULT_FOLLOWED_TAGS,
```
akinwale: For this, I don't think we should override with `DEFAULT_FOLLOWED_TAGS` if no tags are returned by the remote saved state. That may override the local state if the user has made any changes, or in cases where the shared state may have ended up out of sync.

src/redux/actions/sync.js
```diff
@@ -0,0 +1,30 @@
...
+type v0Data = {
+  version: '0.1',
+  shared: {
+    subscriptions?: Array<string>,
+    tags?: Array<string>,
+  },
+};
+
+function extractUserState(rawObj: v0Data) {
+  if (rawObj && rawObj.version === '0.1' && rawObj.shared) {
```
akinwale: Should we be checking for version here? That means we[ll possibly have to update this with every new version, but I suppose it's not too big of a deal since we'll eventually transition to sync using the wallet servers.

seanyesmunt: I think so. With new versions we would have to update anyway since that would probably mean changing the shape of persisted object.

src/redux/reducers/tags.js
seanyesmunt: Maybe 
`if (!followedTags && !tags) return DEFAULT_FOLLOWED)TAGS`
`else return tags`
?

seanyesmunt:
@akinwale I'm going to merge this just so it's easier to work with. We can continue to make changes  to it.

PR #198, opened 2019-09-11
modifies doQuery to take custom types [WIP]

jessopb:
Currently, doSearch related content doesn't take any hints as to file type.


seanyesmunt:
I think this direction looks good.

seanyesmunt:
Reopen if this will be finished

PR #199, opened 2019-09-12
support for ytsync transfers

jessopb:


src/redux/reducers/claims.js
```diff
@@ -302,6 +304,14 @@ reducers[ACTIONS.UPDATE_CHANNEL_COMPLETED] = (state: State, action: any): State
   });
 };
 
+reducers[ACTIONS.IMPORT_CHANNEL_STARTED] = (state: State, action: any): State => {
+  const channelId: string = action.data.id;
+  return Object.assign({}, state, { pendingChannelImports: channelId });
```
seanyesmunt: Since this can only ever be one pending import, this can just be `pendingImport` and we also won't need to pass in the channel id.

PR #200, opened 2019-09-23
transaction pagination selectors

jessopb:
does transaction filtering and pagination in selectors

### DIFF
--- a/src/redux/selectors/wallet.js
+++ b/src/redux/selectors/wallet.js
@@ -2,7 +2,7 @@
 
 import { createSelector } from 'reselect';
 import * as TRANSACTIONS from 'constants/transaction_types';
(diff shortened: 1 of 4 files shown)


src/redux/selectors/wallet.js
```diff
@@ -1,5 +1,8 @@
+// @flow
+
 import { createSelector } from 'reselect';
 import * as TRANSACTIONS from 'constants/transaction_types';
+import { PAGE_SIZE } from 'constants/claim';
```
seanyesmunt: We should use a different number for this. 20 is pretty small for the list. Maybe 50?

jessopb: changed to a local constant. I suppose eventually it should be a setting.

PR #203, opened 2019-09-23
feat: granular balances

tzarebczan:
+ We were not using the total balances stuff I removed -that was an attempt by Akin at one point to work around the SDK cross account balance issue.
+ Decided to keep `balance` as is so we don't have to change it everywhere else...still makes sense I think.


seanyesmunt:
This is great! Nice work.

tzarebczan:
This should be backwards compatible as well...kept balance the same as what we were using before.

seanyesmunt:
👍 can you rebase this branch then merge it?

PR #204, opened 2019-09-24
fix channel list reducer

zxawry:
unsure if it is intentional but reducer for `FETCH_CHANNEL_LIST_COMPLETED` updates `myClaims` with an array of channel claims and removes other stream claims. this causes the related selectors (e.g. `makeSelectClaimIsMine`) produce wrong outputs after any calls to `doFetchChannelListMine`.


akinwale:
Closing this in favour of https://github.com/lbryio/lbry-redux/pull/205/files.

PR #209, opened 2019-09-27
Tv publish api

jessopb:
Changes sdk api call if method=publish and daemonConnectionString contains "proxy" (interested in better way to do this).

### DIFF
--- a/src/constants/headers.js
+++ b/src/constants/headers.js
@@ -1,1 +1,0 @@
-export const AUTH_TOKEN = 'X-Lbry-Auth-Token';
--- a/src/constants/headers.js
+++ b/src/constants/headers.js
@@ -1,0 +1,1 @@
+export const AUTH_TOKEN = 'X-Lbry-Auth-Token';
--- a/src/constants/speech_urls.js
+++ b/src/constants/speech_urls.js
@@ -1,0 +1,2 @@
+export const SPEECH_STATUS = 'https://spee.ch/api/config/site/publishing';
+export const SPEECH_PUBLISH = 'https://spee.ch/api/claim/publish';
--- a/src/lbry.js
+++ b/src/lbry.js
@@ -30,6 +30,7 @@ const Lbry: LbryTypes = {
   setOverride: (methodName, newMethod) => {
     Lbry.overrides[methodName] = newMethod;
   },
+  getApiRequestHeaders: () => Lbry.apiRequestHeaders,
 
   // Returns a human readable media type based on the content type or extension of a file that is returned by the sdk
   getMediaType: (contentType?: string, fileName: ?string) => {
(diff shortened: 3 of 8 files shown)


src/constants/speech_urls.js
```diff
@@ -0,0 +1,2 @@
+export const SPEECH_STATUS = 'https://spee.ch/api/config/site/publishing';
+export const SPEECH_PUBLISH = 'https://spee.ch/api/claim/publish';
```
kauffj: :clap: good change

src/lbry.js
```diff
@@ -160,16 +160,59 @@ function checkAndParse(response) {
 
 function apiCall(method: string, params: ?{}, resolve: Function, reject: Function) {
   const counter = new Date().getTime();
-  const options = {
-    method: 'POST',
-    headers: Lbry.apiRequestHeaders,
-    body: JSON.stringify({
+
+  /*
+    For LBRYTV publishes,
```
kauffj: lbry-redux should not know about lbry.tv

I would find any of the following preferable:
- A distinct method call for publishing, separate from `apiCall`, that has a parameter for local vs. remote (or a boolean for whether to post the file, or equivalent).
- An `@if TARGET` statement by-passing the traditional publish call to so something special for lbry.tv directly in `lbry-desktop`

src/lbry.js
```diff
@@ -160,16 +160,59 @@ function checkAndParse(response) {
...
+    headers: { 'X-Lbry-Auth-Token': sOmEtOken }
+    body.file = the file
+    body.json_payload = { // normal sdk publish json
+      jsonrpc: '2.0',
+      method,
+      params,
+    }
+   */
+  let options;
+  if (method === 'publish' && Lbry.daemonConnectionString.includes('proxy')) {
```
kauffj: this is a magic condition

src/lbry.js
```diff
@@ -160,16 +160,59 @@ function checkAndParse(response) {
...
+    const fileField = params.file_path;
+    // Putting a dummy value here, the server is going to process the POSTed file
+    // and set the file_path itself
+    params.file_path = '__POST_FILE__';
+
+    // FormData() sets the Content-Type to multipart automatically.
+    // We just need the auth token:
+    const token = Lbry.apiRequestHeaders['X-Lbry-Auth-Token'];
```
kauffj: I believe `x-lbry-auth-token` should not be allowed in lbry-redux

src/lbry.js
```diff
@@ -142,6 +142,15 @@ const Lbry: LbryTypes = {
   },
 };
 
+Lbry.publish = (params = {}) =>
```
kauffj: I'm not against this, but what's the point of the `overrides` if we have to write this function?

src/constants/headers.js
```diff
@@ -0,0 +1 @@
+export const AUTH_TOKEN = 'X-Lbry-Auth-Token';
```
kauffj: does this have to be in `lbry-redux`?

PR #211, opened 2019-09-30
shared user state with preferences_get and preferences_set

akinwale:


dist/flow-typed/Lbry.js
```diff
@@ -202,6 +202,10 @@ declare type LbryTypes = {
   blob_delete: (params: {}) => Promise<string>,
   blob_list: (params: {}) => Promise<BlobListResponse>,
 
+  // Preferences
+  preference_get: (params: {}) => Promise<any>,
+  preference_set: (params: {}) => Promise<any>,
```
seanyesmunt: Can you create return types for this?

PR #212, opened 2019-09-30
add commas to formatCredits

seanyesmunt:
Also looks like some additional bundle stuff that wasn't updated in previous PR's

I also added a fix for pending channels


kauffj:
@seanyesmunt it still seems wrong to me that `formatCredits(1000000.000001, 6)` would output the entire number. I think it is better to do something like lower the precision when using a suffix.

seanyesmunt:
Yeah, that makes sense. I'll update it

PR #219, opened 2019-10-14
changes needed for preference sync

seanyesmunt:
- now using middleware to handle calling `preference_set`
- changed ACCOUNT_XXX to WALLET_XXX
  - There are still more redux changes for the wallet commands but those can be addressed later
- Removed `accountId`, and `walletId` params since they aren't being used anywhere.


seanyesmunt:
@akinwale I'd appreciate feedback on the middleware. It requires a little more setup but I think it's a lot easier to track when the call is being made.

The setup changes I had to make are here
https://github.com/lbryio/lbry-desktop/pull/3021/files#diff-7067b67c061ed722ba03e83a0d166738R111-R136

PR #223, opened 2019-10-15
myClaims updated on changes

jessopb:
Please test - abandons seem to work and decrement size of myclaims by 1, and a tip unlock still worked instantly. a channel abandon also decremented myclaims size by 1.


seanyesmunt:
@jessopb I think we might just need to create a new copy with `Object.assign` `slice`

jessopb:
Closed for #225

PR #228, opened 2019-10-21
Changing Pagination count to 10 instead of 50

blokboy:
## PR Checklist

Please check all that apply to this PR using "x":

- [x] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [x] I have checked that this PR does not introduce a breaking change
- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting)
- [x] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other - Please describe:

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/3045

## What is the current behavior?
- The page size for the pagination is divided by TX_LIST.PAGE_SIZE which is currently set to 50, so we have 50 items per page.

## What is the new behavior?
- The application will now display 10 items per page since the Pagination component chooses how many elements to put on each page with the constant provided via Redux.

### DIFF
--- a/src/constants/transaction_list.js
+++ b/src/constants/transaction_list.js
@@ -1,2 +1,2 @@
 // PAGE SIZE
-export const PAGE_SIZE = 50;
+export const PAGE_SIZE = 10;


seanyesmunt:
We shouldn't change that value because it will affect the other list of transactions. This function will need to change https://github.com/lbryio/lbry-redux/blob/master/src/redux/selectors/wallet.js#L218-L231

To just return `x` amount of transactions, regardless of the date.

seanyesmunt:
I'm going to close this. Feel free to re-open if you would still like to try and make this change!

PR #240, opened 2019-11-01
support for 0.45 sdk

seanyesmunt:


seanyesmunt:
@akinwale I don't think there are any breaking changes with this. We're manually passing pages to make some of the calls faster for lbry.tv.

src/redux/actions/publish.js
```diff
@@ -366,8 +370,9 @@ export const doCheckPendingPublishes = (onConfirmed: Function) => (
   let publishCheckInterval;
 
   const checkFileList = () => {
-    Lbry.claim_list().then(claims => {
-      // $FlowFixMe
+    Lbry.stream_list({ page: 1, page_size: 10 }).then(result => {
```
akinwale: What happens here if the stream you want to check isn't on the first page? Does it return the newest item as the first result by default?

tzarebczan: Definitely true for new publishes, and I'll double check for updates.

tzarebczan: Yep, updates are at the top of stream list: `C:\Users\thoma\Desktop>lbrynet stream list --page=1 --page_size=2
{
  "items": [
    {
      "address": "bLX4AyGuGVwdThmRTibNFWQkbWtQEcTWqD",
      "amount": "0.1",
      "claim_id": "c371eed18f74cfd57771838b88dbca602a464ddf",
      "claim_op": "update",
      "confirmations": 0,
      "height": 0,
      "is_change": false,
      "is_channel_signature_valid": true,
      "is_mine": true,
      "meta": {},
      "name": "follow-1",
      "normalized_name": "follow-1",`

PR #250, opened 2019-12-11
Wallet servers

jessopb:


src/constants/shared_prefs.js
```diff
@@ -0,0 +1,3 @@
+import * as DAEMON_SETTINGS from './daemon_settings';
```
seanyesmunt: I don't think this file needs to exist. Just import from `constants/daemon_settings`

src/redux/actions/sync.js
```diff
@@ -8,27 +8,29 @@ type SharedData = {
...
   return {};
 }
 
-export function doPopulateSharedUserState(settings: any) {
+export function doPopulateSharedUserState(sharedSettings: any) {
   return (dispatch: Dispatch) => {
-    const { subscriptions, tags } = extractUserState(settings);
-    dispatch({ type: ACTIONS.USER_STATE_POPULATE, data: { subscriptions, tags } });
+    const { subscriptions, tags, blockedChannels, settings } = extractUserState(sharedSettings);
+    dispatch({ type: ACTIONS.USER_STATE_POPULATE, data: { subscriptions, tags, blockedChannels, settings } }); // clientSettings ? hideSplash ?
```
seanyesmunt: comment not needed

src/redux/actions/wallet.js
```diff
@@ -338,6 +338,20 @@ export function doWalletLock() {
   };
 }
 
+export function doWalletReconnect() {
+  return dispatch => {
+    dispatch({
+      type: ACTIONS.WALLET_RESTART,
+    });
+    // this basically returns null when it's done. :(
+    // might be good to  dispatch ACTIONS.WALLET_RESTARTED
```
seanyesmunt: Can you create an issue against https://github.com/lbryio/lbry-sdk asking for better responses and include a comment for that here?

seanyesmunt: Also

`.then(() => dispatch(...`

instead of 

`.then(dispatch(`

src/constants/shared_prefs.js
jessopb: The idea was, any time daemonSetting is set, if its key is in SHARED_PREFS, we have the middleware save the preference before sending it along to the sharedPreference state. That allows us to pick and choose daemonsettings to keep as preferences.
https://github.com/lbryio/lbry-desktop/blob/f3612084885beb7afce96c5aa2a5093042c6e706/ui/redux/actions/settings.js#L61

PR #251, opened 2019-12-13
fix bug of reporting 0 total claims

jessopb:


dist/bundle.es.js
```diff
@@ -2042,7 +2042,7 @@ const makeSelectNsfwCountForChannel = uri => reselect.createSelector(selectClaim
 });
 
 const makeSelectOmittedCountForChannel = uri => reselect.createSelector(makeSelectTotalItemsForChannel(uri), makeSelectTotalClaimsInChannelSearch(uri), (claimsInChannel, claimsInSearch) => {
-  if (claimsInChannel && claimsInSearch) {
+  if (claimsInChannel && typeof claimsInSearch === 'number' && claimsInSearch >= 0) {
```
jessopb: apparently null >= 0. so the typeof is probably a good defensive move

PR #256, opened 2019-12-18
pass related_to flag to search for recommended content

seanyesmunt:
https://lbryians.slack.com/archives/C1R07H02X/p1576641588024400?thread_ts=1576637646.019900&cid=C1R07H02X

I think we should just get rid of the search caching, it makes everything so complex. I'd like to do some cleanup when I start working on it more app side, but this should do for now. Passing `related_to={claim_id}` is needed to improve related content a lot.

### DIFF
--- a/dist/bundle.es.js
+++ b/dist/bundle.es.js
@@ -2202,7 +2202,7 @@ const makeSelectRecommendedContentForUri = uri => reselect.createSelector(makeSe
       return;
     }
 
-    const searchQuery = getSearchQueryString(title, undefined, undefined, {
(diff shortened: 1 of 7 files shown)


seanyesmunt:
@akinwale With this, wherever you call `doSearch` with the title from a file, you will have to pass `{ related_to: {claim_id} }` for this to work.

Desktop changes
https://github.com/lbryio/lbry-desktop/pull/3398/files

seanyesmunt:
also snuck in a couple other fixes for the desktop app release

PR #258, opened 2020-01-05
add doResolvedSearch actions which returns resolved search results

akinwale:


akinwale:
I separated this since `doSearch` already had its own resolve logic, and I didn't want to make the entire thing more complicated than necessary.

src/redux/actions/search.js
```diff
@@ -159,6 +160,77 @@ export const doSearch = (
     });
 };
 
+export const doResolvedSearch = (
```
kauffj: Do we really have to do this with an entirely parallel chain of calls, as well as a new `ResolvedSearchResult` type? Why can't this be an additional parameter to `doSearch`, and when pre-resolved results are returned from the search server, they're stored directly in the claims reducer?

(I realize that part of this is that `resolve=1` on lighthouse doesn't actually return the resolved claim, which makes me question the design of this entirely, or at a minimum the name of that parameter)

akinwale: The reasoning behind this approach was to have a type with the minimum number of fields required to display a search result item, since I was aiming for speed and better performance (one request vs multiple requests to handle a search). The performance improvement on mobile is positive.

I created a new method instead of using the existing `doSearch` because I was trying to differentiate the return types (since it ultimately doesn't return a claim, but a subset of values for a particular claim). 

@tiger5226 could probably rename the `resolve` parameter to something else to make it more clear.

tiger5226: I can change it to whatever we want it to be.

PR #259, opened 2020-01-07
Updates comment-related redux code to support sdk version 0.53.0

osilkin98:
Adds support for LBRY SDK Version `0.53.0`. 

### Adds
Three new functions are added to `Lbry.js` to communicate directly with the SDK: 
- `comment_hide`
- `comment_abandon`
- `comment_update`

Similarly, three new functions are added to `src/redux/actions/comments.js`: 
1. `doCommentHide(comment_id: String)`
2. `doCommentAbandon(comment_id: string)`
3. `doCommentUpdate(comment_id: string, comment: string)`
 

### Changes:
`state.comments.byId` now maps a `claim_id` to a `comment_id` list
`state.comments.commentById` is added to map a `comment_id` to a `Comment` object
`state.comments.commentsByUri` maps a `uri` to a `claim_id`

`type Comment` is modified to be reflective of how the API retrieves comments


src/redux/actions/comments.js
```diff
@@ -82,3 +83,98 @@ export function doCommentCreate(
...
+        });
+      })
+      .catch(error => {
+        dispatch({
+          type: ACTIONS.COMMENT_HIDE_FAILED,
+          data: error,
+        });
+        dispatch(
+          doToast({
+            message: `SDK Errored when trying to hide Comment with comment_id: "${comment_id}"`,
```
seanyesmunt: Users don't care about comment_id's. Maybe something like this:

"There was an error hiding this comment. Please try again later."

src/redux/actions/comments.js
```diff
@@ -82,3 +83,98 @@ export function doCommentCreate(
...
+        });
+      })
+      .catch(error => {
+        dispatch({
+          type: ACTIONS.COMMENT_ABANDON_FAILED,
+          data: error,
+        });
+        dispatch(
+          doToast({
+            message: `SDK Errored during abandon on Comment w/ ID = "${comment_id}"`,
```
seanyesmunt: Same here.

src/redux/actions/comments.js
```diff
@@ -82,3 +83,98 @@ export function doCommentCreate(
...
+            isError: true,
+          })
+        );
+      });
+  };
+}
+
+export function doCommentUpdate(comment_id: string, comment: string) {
+  // if they provided an empty string, they must have wanted to abandon
+  if (comment === '') {
```
seanyesmunt: Nice. I like this pattern

src/redux/actions/comments.js
```diff
@@ -82,3 +83,98 @@ export function doCommentCreate(
...
+            data: {
+              comment: result,
+            },
+          });
+        })
+        .catch(error => {
+          dispatch({ type: ACTIONS.COMMENT_UPDATE_FAILED, data: error });
+          dispatch(
+            doToast({
+              message: `SDK Errored during update on Comment w/ ID = ${comment_id}`,
```
seanyesmunt: Same here

src/redux/reducers/comments.js
```diff
@@ -58,6 +81,79 @@ export const commentReducer = handleActions(
...
+      const commentById = Object.assign({}, state.commentById);
+      const byId = Object.assign({}, state.byId);
+
+      if (abandoned && comment_id in abandoned) {
+        // messy but necessary for the time being
+        const comment: Comment = commentById[comment_id];
+        const commentIds = byId[comment.claim_id];
+        byId[comment.claim_id] = commentIds.filter(commentId => commentId !== comment_id);
+
+        Object.keys(commentById).forEach(commentId => {
```
seanyesmunt: I think this part can just be replaced with 
```
delete commentById[comment_id]
```

seanyesmunt: No need to loop through it.

src/redux/selectors/comments.js
```diff
@@ -5,9 +5,37 @@ const selectState = state => state.comments || {};
...
+      // map a new array of comments by the claimId
+      comments[claimId] = Array(commentIds === null ? 0 : commentIds.length);
+      for (let i = 0; i < commentIds.length; i++) {
+        comments[claimId][i] = byId[commentIds[i]];
+      }
+    });
+
+    return comments;
+  }
 );
```
seanyesmunt: Why do we need this new selector? (Not saying we don't, I'm just trying to understand the logic)

Is it that in you first get a list of the children comment ids, then for each child, get the actual comment?

It also helps to explain why you are doing something in a comment instead of what the code is doing, since that can usually be figured out. 

ex: "map a new array of comments by the claimId"

I can see that's what this code is doing, but I don't know why it's doing that.

PR #260, opened 2020-01-15
Fix wrong language when editing claim

Yamboy1:
Fixes lbryio/lbry-desktop#3479

Also see lbryio/lbry-desktop#3481

Note that this is my first time using redux, so i'm sorry if it's a bit unconventional :sweat_smile:.


seanyesmunt:
Nice! I just made one small change and moved the PR to here https://github.com/lbryio/lbry-redux/pull/261

I will merge that and close this one.

PR #271, opened 2020-02-06
basic repost support

seanyesmunt:
Also changed `doClaimListMine` to call `claim_list` with `claim_type = ['stream', 'repost']`


seanyesmunt:
@kauffj says 🚢

PR #275, opened 2020-02-12
add repost_url and also add a way to return the original claim instead of the reposted claim

seanyesmunt:


seanyesmunt:
Merging this since it fixes something breaking on lbry.tv currently, but unsure if this is the best approach

PR #280, opened 2020-02-25
Enforce channel comments

osilkin98:
Sets the `Channel` param as mandatory in `commentCreate`, and provides a toast for the situation.


seanyesmunt:
@osilkin98 sorry this has been sitting for so long. I'll review/merge today

PR #287, opened 2020-03-12
light tag cleanup

kauffj:


src/constants/tags.js
```diff
@@ -239,7 +228,6 @@ export const DEFAULT_KNOWN_TAGS = [
   'dota 2',
   'studio',
   'star wars',
-  'gta 5',
```
seanyesmunt: Why keep `dota 2` above but get rid of `gta 5`?

src/constants/tags.js
```diff
@@ -411,7 +395,6 @@ export const DEFAULT_KNOWN_TAGS = [
   'style',
   'travel trailer',
   'rda',
-  '5859dfec-026f-46ba-bea0-02bf43aa1a6f',
```
seanyesmunt: nice

src/constants/tags.js
kauffj: I kept `gta v` instead

PR #292, opened 2020-03-19
add reposted property

kauffj:


kauffj:
companion to https://github.com/lbryio/lbry-desktop/pull/3862

PR #293, opened 2020-03-25
tip management on claim

jessopb:


src/redux/selectors/wallet.js
```diff
@@ -21,6 +21,24 @@ export const selectWalletEncryptSucceeded = createSelector(
...
+  selectState,
+  state => state.pendingSupportTransactions
+);
+
+export const makeSelectPendingAmountByUri = (uri) => createSelector(
+  selectClaimIdsByUri,
+  selectPendingSupportTransactions,
+  (claimIdsByUri, pendingSupports) => {
+    // claimIdsByUri: { uristring: claimIdString, ....}
+    // pendingSupports: { id: { txid, effectiveamount }, ... }
```
seanyesmunt: comment

src/redux/actions/wallet.js
```diff
@@ -413,3 +439,49 @@ export function doUpdateBlockHeight() {
...
+      dispatch(doFetchClaimListMine());
+      dispatch(doFetchChannelListMine());
+    }
+
+    if (!Object.keys(pendingTxsById).length) {
+      clearInterval(txCheckInterval);
+    }
+  };
+
+  console.log()
```
seanyesmunt: console.log

src/redux/actions/wallet.js
```diff
@@ -413,3 +439,49 @@ export function doUpdateBlockHeight() {
...
+  let changed = false;
+
+  if (!Object.keys(pendingTxsById).length) {
+    return;
+  }
+  let txCheckInterval;
+  const checkTxList = () => {
+    Object.entries(pendingTxsById).forEach(async([claim, data]) => {
+      await Lbry.transaction_show({txid: data.txid}).then(result => {
+        console.log(result);
```
seanyesmunt: console.log

src/redux/actions/wallet.js
```diff
@@ -413,3 +439,49 @@ export function doUpdateBlockHeight() {
...
+  const pendingTxsById = selectPendingSupportTransactions(state); // {}
+  let changed = false;
+
+  if (!Object.keys(pendingTxsById).length) {
+    return;
+  }
+  let txCheckInterval;
+  const checkTxList = () => {
+    Object.entries(pendingTxsById).forEach(async([claim, data]) => {
+      await Lbry.transaction_show({txid: data.txid}).then(result => {
```
seanyesmunt: Since you are using `await` here you can also do 
```
const result = await Lbry.transaction_show();
if (result.height > 0) {
```

Right now it isn't really doing anything

src/redux/actions/wallet.js
```diff
@@ -413,3 +439,49 @@ export function doUpdateBlockHeight() {
...
+    });
+
+    if (changed) {
+      dispatch({
+        type: ACTIONS.PENDING_SUPPORTS_UPDATED,
+        data: {
+          pending: pendingTxsById,
+        },
+      });
+      dispatch(doFetchClaimListMine());
```
seanyesmunt: Is it possible to look at something inside the `transaction_show` call to know if it's a channel or claim? Then we shouldn't need to call both `claim_list` and `channel_list`

src/redux/actions/wallet.js
```diff
@@ -342,6 +343,31 @@ export function doWalletLock() {
   };
 }
 
+export function doSupportAbandonForClaim(claimId, keep, preview = true) {
+  return dispatch => {
+    if (preview === false) {
```
seanyesmunt: When would someone pass `true` for preview?

src/constants/action_types.js
```diff
@@ -39,6 +39,9 @@ export const FETCH_SUPPORTS_STARTED = 'FETCH_SUPPORTS_STARTED';
 export const FETCH_SUPPORTS_COMPLETED = 'FETCH_SUPPORTS_COMPLETED';
 export const ABANDON_SUPPORT_STARTED = 'ABANDON_SUPPORT_STARTED';
 export const ABANDON_SUPPORT_COMPLETED = 'ABANDON_SUPPORT_COMPLETED';
+export const ABANDON_CLAIM_SUPPORT_STARTED = 'ABANDON_CLAIM_SUPPORT_STARTED';
+export const ABANDON_CLAIM_SUPPORT_COMPLETED = 'ABANDON_CLAIM_SUPPORT_COMPLETED';
```
seanyesmunt: Please add `ABANDON_CLAIM_SUPPORT_FAILED` along with a value in the reducer, just to track the error even if we aren't using it at the moment.

src/redux/actions/wallet.js
jessopb: rewrote as Promise.all().then().then()

src/redux/actions/wallet.js
jessopb: I stored the value type in pending and only run channel if it has one.

src/redux/actions/wallet.js
jessopb: changed this a bit.I just wanted it safely disabled during testing.

src/redux/actions/wallet.js
jessopb: now storing claim value type in the pending.

PR #295, opened 2020-03-30
save optimize check state to allow page navigation

jessopb:
This addresses navigating away from publish page after having selected a file. Browser security doesn't make it simple to load a file the same as an <input> does without a user-triggered action.

### DIFF
--- a/src/redux/reducers/publish.js
+++ b/src/redux/reducers/publish.js
@@ -9,6 +9,10 @@ type PublishState = {
   editingURI: ?string,
   filePath: ?string,
   contentIsFree: boolean,
+  canOptimize: ?boolean,
+  checkFileDur: number,
+  checkFileSize: number,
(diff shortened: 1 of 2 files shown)


src/redux/reducers/publish.js
```diff
@@ -9,6 +9,10 @@ type PublishState = {
   editingURI: ?string,
   filePath: ?string,
   contentIsFree: boolean,
+  canOptimize: ?boolean,
+  checkFileDur: number,
+  checkFileSize: number,
+  checkFileVid: boolean,
```
seanyesmunt: I would remove the `check` prefix here.
`fileDuration`
`fileSize`
`fileIsVideo`

seanyesmunt: Unless `check` is referencing something that I don't understand

PR #299, opened 2020-04-10
Feat txo paginate

jessopb:


src/constants/action_types.js
```diff
@@ -35,6 +35,9 @@ export const GET_NEW_ADDRESS_STARTED = 'GET_NEW_ADDRESS_STARTED';
 export const GET_NEW_ADDRESS_COMPLETED = 'GET_NEW_ADDRESS_COMPLETED';
 export const FETCH_TRANSACTIONS_STARTED = 'FETCH_TRANSACTIONS_STARTED';
 export const FETCH_TRANSACTIONS_COMPLETED = 'FETCH_TRANSACTIONS_COMPLETED';
```
seanyesmunt, line 37: Do you need these old actions still?

src/index.js
```diff
@@ -106,6 +106,7 @@ export {
   doUpdateBalance,
   doBalanceSubscribe,
   doFetchTransactions,
```
seanyesmunt, line 108: and this? Can we nuke it all?

src/redux/actions/claims.js
```diff
@@ -183,13 +183,7 @@ export function doAbandonClaim(txid: string, nout: number) {
...
         })
       );
-
-      // After abandoning, fetch transactions to show the new abandon transaction
-      // Only fetch the latest few transactions since we don't care about old ones
-      // Not very robust, but better than calling the entire list for large wallets
-      const page = 1;
-      const pageSize = 10;
-      dispatch(doFetchTransactions(page, pageSize));
+      // we used to fetch recent transactions into the full list here.
```
seanyesmunt, line 186: can we call txo_list for page 1 here?

src/redux/actions/wallet.js
```diff
@@ -72,6 +72,28 @@ export function doFetchTransactions(page = 1, pageSize = 99999) {
   };
 }
 
+export function doFetchTransactionPage(queryParams) {
```
seanyesmunt, line 75: Please add a type for `TXO_LIST_PARAMS`

src/redux/selectors/wallet.js
```diff
@@ -321,6 +321,26 @@ export const selectFilteredTransactions = createSelector(
   }
 );
 
+export const selectTransactionPage = createSelector(
+  selectState,
+  state => state.transactionPage ? state.transactionPage.items || [] : [],
```
seanyesmunt, line 326: this could be `(state.transactionPage && state.transactionPage.items) || []`

src/redux/selectors/wallet.js
```diff
@@ -321,6 +321,26 @@ export const selectFilteredTransactions = createSelector(
...
+);
+
+export const selectTransactionPageNumber = createSelector(
+  selectState,
+  state => state.transactionPage ? state.transactionPage.page || 1 : 1,
+);
+
+export const selectTransactionItemCount = createSelector(
+  selectState,
+  state => state.transactionPage ? state.transactionPage.total_items || 1 : 1,
```
seanyesmunt, line 336: same with these too. Since you are doing `|| 1 : 1`, it could just be `value || 1`

src/constants/action_types.js
jessopb: Desktop still has access to the original transaction list.

src/redux/actions/claims.js
jessopb: Considered it. 
Currently the component parses the params locally and feeds them to fetchTransactoins(p).
Instead, we would have to parse the params locally and update redux with the params, 'dot then' fetch given the params.
Then we could ask redux to refetch current params at any time...

src/constants/action_types.js
jessopb: The old page brings only the file export of the list. EVentually I figure that would be a modal.

src/redux/actions/wallet.js
jessopb: resolved to not enable @flow for this file at this time.

src/redux/actions/claims.js
jessopb: Changed to updating params which calls fetchTxos on success.

src/index.js
jessopb: Not if we're reimplementing the exporter.

src/redux/actions/claims.js
```diff
@@ -120,6 +121,91 @@ export function doFetchClaimListMine(
...
+
+    const successCallback = () => {
+      dispatch({
+        type: completedActionType,
+        data,
+      });
+
+      let abandonMessage;
+      if (isClaim) {
+        abandonMessage = 'Successfully abandoned your claim.';
```
seanyesmunt, line 163: i18n

PR #306, opened 2020-04-24
fetch claim list by page

jessopb:


src/redux/actions/claims.js
```diff
@@ -35,6 +35,11 @@ export function doResolveUris(uris: Array<string>, returnCachedClaims: boolean =
       return;
     }
 
+    const options = {};
```
seanyesmunt, line 38: this should be
`const options: { include_is_my_output?: boolean } = {};`

You might have to change the way you add it below to 
`options.include_is_my_output = true`

src/redux/actions/claims.js
```diff
@@ -583,6 +594,30 @@ export function doRepost(options: StreamRepostOptions) {
...
+  return (dispatch: Dispatch) => {
+    dispatch({
+      type: ACTIONS.CHECK_PUBLISH_NAME_STARTED,
+    });
+
+    return Lbry.claim_list({ name: name }).then(result => {
+      dispatch({
+        type: ACTIONS.CHECK_PUBLISH_NAME_COMPLETED,
+      });
+      if (result.items.length) {
```
jessopb, line 607: @seanyesmunt added this section to fix publish edits. Every potential name we check for availability, if found, we add to my claims. this makes the name checking component work.

seanyesmunt: Makse sense 👍

PR #313, opened 2020-05-07
Support lbry:// links with timestamp

jeffslofish:
Current behavior:
If you paste a lbry:// link with a ?t=xyz parameter at the end into the wunderbar, it will start the video at the beginning.

New behavior:
If you paste a lbry:// link with a ?t=xxx parameter at the end into the wunderbar, it will start the video at the xyz seconds timestamp.

This should fix https://github.com/lbryio/lbry-desktop/issues/3972


seanyesmunt:
Sorry this took so long to merge. I missed this one. This is awesome! 🙂

PR #314, opened 2020-05-11
Publish mode

btzr-io:
Identify how the user is publishing the content:
- Input file ( local file system )
- Drag-drop file ( local file system ) https://github.com/lbryio/lbry-desktop/issues/350
- In-app text / markdown generated file ( web fileSystem API ) https://github.com/lbryio/lbry-desktop/issues/4105
- Others ?....

```js
// OnFileDrop:
updatePublishForm({ filePath: string || webFile, mode: PUBLISH_MODE.DROP } });

// Select file on fileSelector:
// OnFileChange:
updatePublishForm({ filePath: string || webFile, mode: PUBLISH_MODE.INPUT } });

// Post creation
const post = createFile().write("text from user input");
updatePublishForm({ filePath: post, mode: PUBLISH_MODE.POST } });
```

### Todo
- [ ]  Constant types for publish modes

### DIFF
--- a/src/redux/reducers/publish.js
+++ b/src/redux/reducers/publish.js
@@ -6,6 +6,7 @@ import * as THUMBNAIL_STATUSES from 'constants/thumbnail_upload_statuses';
 import { CHANNEL_ANONYMOUS } from 'constants/claim';
 
 type PublishState = {
+  mode: ?string,
   editingURI: ?string,
   filePath: ?string,
   contentIsFree: boolean,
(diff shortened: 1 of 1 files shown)


btzr-io:
I think this approach is really simple, not sure if is the best to handle new features / forms of publishing ?

seanyesmunt:
Does the UI change in "drop" mode? Or does that just make it so dropping a file works?

btzr-io:
Nope it does not change the UI, I think I can make it work without this but I'll need to rewrite more code...

Publish mode will probably change the UI, but I don't know yet how exactly will it work.

seanyesmunt:
👍 If it's not changing the UI I don't think it will need it's own mode?

PR #317, opened 2020-05-26
add case for when amount is undefined in formatCredits instead of returning 0

clay53:
Step towards resolving lbryio/lbry-desktop#4140


clay53:
This is required for the change in function for PR lbryio/lbry-desktop#4247

seanyesmunt:
Awesome!

PR #318, opened 2020-05-26
fixed regression where 0 amounts were getting formatted as -- in formatCredits

clay53:
Sorry, that was my fault. Bug came from #317


seanyesmunt:
I ended up making a change in the lbry-desktop repo. I think it makes more sense to let the components handle what they should display so I am going to close this.

PR #321, opened 2020-06-03
Incorporating default tags in spanish

michaelmitnick:
It is my first fork and pull request, please be kind. I hope the task is correct and helps LBRY. I did not implement a new constant because they told me to only add the labels in the DEFAULT_KNOWN_TAGS constant that was already created.


src/constants/tags.js
```diff
@@ -520,4 +520,21 @@ export const DEFAULT_KNOWN_TAGS = [
   '2020protests',
   'covidcuts',
   'covid-19',
+  'español',
+  'tecnología',
+  'criptomonedas',
+  'economía',
+  'bitcoin',
```
kauffj, line 527: @seanyesmunt I told @michaelmitnick to do this but we should probably split the Spanish tags into a separate array for easier future refactoring/cleanup

kauffj: (and call `.unique()` on resultant merge)

seanyesmunt:
@michaelmitnick I made a small change and pushed it up to your branch

PR #323, opened 2020-06-16
Update strings for translation

TigerxWood:


TigerxWood:
And that linktarget from line 269 should not be translated right?

seanyesmunt:
Yep. Good catch!

PR #324, opened 2020-06-17
Update bundle.es.js with Romanian default tags

TigerxWood:


seanyesmunt:
Hey @TigerxWood 

We are only supporting spanish tags for now to work with lbry.lat

PR #329, opened 2020-06-24
Fix unresolved "You sent ${amount} LBC"

infinite-persistence:
## Fixes:
[lbry-desktop] [4429](https://github.com/lbryio/lbry-desktop/issues/4429): `Unresolved string: "You sent ${amount} LBC"`

## Changes:
Corrected error in commit c9492522.

### DIFF
--- a/src/redux/actions/wallet.js
+++ b/src/redux/actions/wallet.js
@@ -187,7 +187,7 @@ export function doSendDraftTransaction(address, amount) {
         });
         dispatch(
           doToast({
-            message: __('You sent ${amount} LBC'),
+            message: __(`You sent ${amount} LBC`),
             linkText: __('History'),
             linkTarget: '/wallet',
           })


TigerxWood:
Are you sure that is the problem? I just updated the strings in #322 for translation, but the strings are not yet in the file for translations. Because if the cause is of the problem is using ( ' ) instead of ( ` ) then many strings will have this problem. I will update soon the app-strings.jsx with all the missing strings from lbry-desktop and lbry-redux and I hope this problem will be solved.

infinite-persistence:
Yes.  Totally different scenario.
- Read: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
- Backticks are not needed for `%xxx%` type of substitution used in `<i18nMessage>` tags because the `__()` macro is doing the substitution instead.

@sean, FYI.

TigerxWood:
Yes, you are right.  I introduced that (') by mistake when I changed it for localisation.

PR #330, opened 2020-06-27
Add file text to publish state

btzr-io:
> Edit previous text or markdown posts directly in the app

Added the contents of the file as string if contentType is "text", this value will be used for the markdown editor
Required by In-app text and markdown publishing and editing [#4105](https://github.com/lbryio/lbry-desktop/issues/4105)


btzr-io:
Now, I'm not really sure if I need this, I could just use the local state of the component, but this will make more easy to clear the text area / editor on publish cancel.

seanyesmunt:
I think since all of the other publish stuff is in redux, it makes sense to keep this here.

PR #332, opened 2020-07-01
Some new translation strings

TigerxWood:


src/redux/actions/notifications.js
```diff
@@ -4,7 +4,7 @@ import uuid from 'uuid/v4';
 
 export function doToast(params: ToastParams) {
   if (!params) {
-    throw Error("'params' object is required to create a toast notification");
+    throw Error(__("'params' object is required to create a toast notification"));
```
seanyesmunt, line 7: this doesn't need translation since it's just a message for developers

src/redux/actions/claims.js
```diff
@@ -231,7 +231,7 @@ export function doAbandonClaim(txid: string, nout: number, cb: string => void) {
     const supportToAbandon = mySupports[outpoint];
 
     if (!claimToAbandon && !supportToAbandon) {
-      console.error('No associated support or claim with txid: ', txid);
+      console.error(__('No associated support or claim with txid: '), txid);
```
seanyesmunt, line 234: the `console.error` messages don't need to be translated.

seanyesmunt:
@TigerxWood I reverted this because it was causing issues with the electron app starting. I'm not sure why that would happen. I also realized this string is setup incorrectly:
https://github.com/lbryio/lbry-redux/pull/332/files#diff-ef4b0f7b9d25dc7919a6d7677fba0e85R40

it should be

```
__("File timeout for %uri%", { uri })
```

If you add this back, can you please test on desktop too?

PR #336, opened 2020-07-10
Publish: Handle reflecting-state not updated correctly.

infinite-persistence:
## Issue
Fixes https://github.com/lbryio/lbry-desktop/issues/4385 `Don't show upload progress if no file exists (shows 0% progress)`

## Assessment
An undefined result exception was causing the reflecting-state update code to be missed, so Desktop still thinks the item is still reflecting and continues to show the `Uploading (0%)` status.

image


seanyesmunt:
Thanks! I'll get this added to lbry-desktop and include it in the release

PR #354, opened 2020-10-13
remove tags from redux (to desktop)

jessopb:


seanyesmunt (changes requested):
Please delete `redux/{actions,reducers,selectors}/tags` too

PR #365, opened 2020-11-06
claim_search: Don't clear previous page results if subsequent pages timeout.

infinite-persistence:
## Issue
https://github.com/lbryio/lbry-desktop/issues/4609

## Change
- Don't clear existing results on timeout.
- Mark the "no more results" flag so that the app knows it needs to stop querying.

### DIFF
--- a/src/redux/reducers/claims.js
+++ b/src/redux/reducers/claims.js
@@ -570,13 +570,20 @@ reducers[ACTIONS.CLAIM_SEARCH_FAILED] = (state: State, action: any): State => {
   const { query } = action.data;
   const claimSearchByQuery = Object.assign({}, state.claimSearchByQuery);
   const fetchingClaimSearchByQuery = Object.assign({}, state.fetchingClaimSearchByQuery);
+  const claimSearchByQueryLastPageReached = Object.assign({}, state.claimSearchByQueryLastPageReached);
 
   delete fetchingClaimSearchByQuery[query];
-  claimSearchByQuery[query] = null;
+
+  if (claimSearchByQuery[query] && claimSearchByQuery[query].length !== 0) {
+    claimSearchByQueryLastPageReached[query] = true;
+  } else {
+    claimSearchByQuery[query] = null;
+  }
 
   return Object.assign({}, state, {
     fetchingClaimSearchByQuery,
     claimSearchByQuery,
+    claimSearchByQueryLastPageReached,
   });
 };
 
--- a/dist/bundle.es.js
+++ b/dist/bundle.es.js
@@ -5155,13 +5155,20 @@ reducers[CLAIM_SEARCH_FAILED] = (state, action) => {
   const { query } = action.data;
   const claimSearchByQuery = Object.assign({}, state.claimSearchByQuery);
   const fetchingClaimSearchByQuery = Object.assign({}, state.fetchingClaimSearchByQuery);
+  const claimSearchByQueryLastPageReached = Object.assign({}, state.claimSearchByQueryLastPageReached);
 
   delete fetchingClaimSearchByQuery[query];
-  claimSearchByQuery[query] = null;
+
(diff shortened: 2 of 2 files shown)


src/redux/reducers/claims.js
```diff
@@ -572,6 +572,22 @@ reducers[ACTIONS.CLAIM_SEARCH_FAILED] = (state: State, action: any): State => {
...
+
+  if (claimSearchByQuery[query] && claimSearchByQuery[query].length !== 0) {
+    const claimSearchByQueryLastPageReached = Object.assign(
+      {},
+      state.claimSearchByQueryLastPageReached
+    );
+
+    claimSearchByQueryLastPageReached[query] = true;
+
+    return Object.assign({}, state, {
```
seanyesmunt, line 584: I would prefer if you didn't do an early return here, and just updated the `claimSearchByQuery` value used in the return below. Currently (I believe) every reducer only has 1 return statement

infinite-persistence:
`Changes`
- Created separate flag to differentiate between "no more results" and "partial-page timeout", as sometimes we'll need to purge the data as `query` doesn't hold the Freshness info.
- Avoid doing early returns as suggested.

src/redux/reducers/claims.js
```diff
@@ -570,13 +581,23 @@ reducers[ACTIONS.CLAIM_SEARCH_FAILED] = (state: State, action: any): State => {
   const { query } = action.data;
   const claimSearchByQuery = Object.assign({}, state.claimSearchByQuery);
   const fetchingClaimSearchByQuery = Object.assign({}, state.fetchingClaimSearchByQuery);
+  const claimSearchByQueryPartialTimeout = Object.assign(
+    {},
+    state.claimSearchByQueryPartialTimeout
+  );
 
   delete fetchingClaimSearchByQuery[query];
-  claimSearchByQuery[query] = null;
```
seanyesmunt, line 575: Sorry I didn't ask this before, what about only setting this to `null` when there are no values already? Then we don't need a new value to track

infinite-persistence: > _what about only setting this to null when there are no values already?_

What does "values" mean in this case?

seanyesmunt: If `claimSeaerchByQuery` isn't cleared out when the call fails, then I don't think we need `claimSearchByQueryPartialTimeout`

```
  if (claimSearchByQuery[query] && claimSearchByQuery[query].length !== 0) {
    // set claimSearchByQueryLastPageReached?
  } else {
    // Only set it to null if there aren't any previous results
    claimSearchByQuery[query] = null;
  }
```

infinite-persistence: > _Then we don't need a new value to track_

claimList.js currently uses const timedOut = uris === null to detect the timed out case. Since it might not be null now, seems like we still need a flag to differentiate between the regular populated case vs partially populated?

seanyesmunt: Good point. I think this is fine then

infinite-persistence: Ah, I missed your reply.  I'll give `claimSearchByQueryLastPageReached` another shot.  I recall I had to differentiate between "last page reached" and "timed out" in the GUI, but it might just be I didn't think deep enough.

seanyesmunt: If there are results already loaded, it's fine not to show the "timed out" message. I can just be treated like the last page was reached

infinite-persistence:
`Change`
- Reverted back to the simple version of just using an existing flag + not purging if results already exists.  The only downside to this is that the "timeout" message will no longer appear, and it might look unresponsive when switching between various filters that all result in timeout.

seanyesmunt:
@infinite-persistence Does the timeout message appear for searches that timeout on the first page?

infinite-persistence:
> @infinite-persistence Does the timeout message appear for searches that timeout on the first page?

Yup, for that case (count = 0), the message will appear as per existing version.

PR #366, opened 2020-11-06
pass languages in channel create

jessopb:


jessopb:
app will require a bump after merge

PR #369, opened 2020-12-05
Select pending url

jessopb:


src/redux/selectors/claims.js
```diff
@@ -383,6 +383,14 @@ export const makeSelectAmountForUri = (uri: string) =>
     }
   );
 
+export const makeSelectEffectiveAmountForUri = (uri: string) =>
+  createSelector(
+    makeSelectClaimForUri(uri),
+    claim => {
+      return claim && claim.meta && claim.meta.effective_amount;
```
seanyesmunt, line 390: I think this should return a number if it exists
```
return claim && claim.meta && Number(claim.meta.effective_amount);
```

jessopb: As Number(undefined) = NaN, I did typeof ..effective_amount === 'string' too.

PR #374, opened 2021-01-04
Utxo counts

jessopb:


jessopb:
I'm satisfied this refactor doesn't break anything. Tested by "unlocking" supports on file page. Main new function is to consolidate utxo payments.

PR #380, opened 2021-02-03
better track txids for large wallet operations

jessopb:
Needed a way to tell which buttons to show or disable during pending state.


src/redux/actions/wallet.js
```diff
@@ -630,6 +636,22 @@ export const doCheckPendingTxs = () => (dispatch, getState) => {
         }
       }
       if (noLongerPendingConsolidate.length) {
+        if (noLongerPendingConsolidate.includes(pendingConsTxid)) {
+          dispatch(
+            doToast({
+              title: __('Wallet Job'),
```
seanyesmunt, line 642: "Ready to go"

seanyesmunt (approved):
one copy change

PR #386, opened 2021-03-05
populate channel claims for txo signing_channels

jessopb:


src/redux/selectors/claims.js
```diff
@@ -94,6 +94,12 @@ export const selectReflectingById = createSelector(
   state => state.reflectingById
 );
 
+export const makeSelectClaimForClaimId = (claimId: string) =>
+  createSelector(
+    selectClaimsById,
+    byId => (byId[claimId] ? byId[claimId] : undefined)
```
seanyesmunt, line 100: this can just be `byId[claimId]` since it will be undefined if it doesn't exist, and also will return `null` correctly for abandoned claims

PR #389, opened 2021-03-14
Persist BTC swap address across devices

infinite-persistence:
## Issue
Used by [lbry-desktop -- Support for swapping into LBC](https://github.com/lbryio/lbry-desktop/pull/5654)


infinite-persistence:
### Update
- Changed from `btcAddresses` to `btc_addresses` to match the others in the wallet.
- Rebased to latest `master`

PR #396, opened 2021-03-18
Drop old txo-fetch results

infinite-persistence:
## Issue
Closes lbry-desktop [4317: Transaction list shows previously requested data / pages](https://github.com/lbryio/lbry-desktop/issues/4317)

## Approach
A naive approach of creating a random transaction ID for each fetch. The latest ID, stored in `state`, will be the expected one -- any other transaction results will be dropped.

The loading spinner will continue to spin until the latest ID's results are fetched.

## Sample
Example of 3 fast filter switches:
image

## ??
This assumes there's no quick, built-in way to abort dispatched tasks (only did a quick Google search for it).  The ability to abort would speed up the wait-time.

### DIFF
--- a/dist/bundle.es.js
+++ b/dist/bundle.es.js
@@ -2840,8 +2840,11 @@ function doFetchTransactions(page = 1, pageSize = 99999) {
 
 function doFetchTxoPage() {
   return (dispatch, getState) => {
+    const fetchId = Math.random().toString(36).substr(2, 9);
+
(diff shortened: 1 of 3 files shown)


jessopb:
currently testing on bob

tzarebczan:
<3 looks great!

jessopb (commented):
:+1:

PR #412, opened 2021-05-11
Change how release_time is edited.

infinite-persistence:
## Issue
For `lbry-desktop #5514 - Add release date field`

## Changes
Updates for  `UpdatePublishFormData`:
- `releaseTime` is now a number instead of a string, matching `release_time`. It was getting confusing what the units were between the app variable vs claim variable.
- `releaseTime` will always match `release_time` for an edit. It will be used in the GUI to reset just the date to the original, instead of having to reset the entire form.
- `releaseTime` will be undefined for new uploads.
- `releaseTimeEdited` will be used by `updatePublishForm` in the GUI to represent the desired new release time. Set to `undefined` if we don't want to change the date.


tzarebczan:
Will give this a test soon, but this doesn't seem right as we always want to populate release time: "releaseTime will be undefined for new uploads."

infinite-persistence:
`releaseTime` and `releaseTimeEdited` is only used in the GUI side to differentiate the following cases:
- creating new claim
- editing claim
- editing claim, and need to get back original/previous release time.
- etc.

Ultimately, they get converted to the following when publishing:
```
  if (releaseTimeEdited) {
    publishPayload.release_time = releaseTimeEdited;
  } else if (myClaimForUriEditing && myClaimForUriEditing.value.release_time) {
    publishPayload.release_time = Number(myClaimForUri.value.release_time);
  } else if (myClaimForUriEditing && myClaimForUriEditing.timestamp) {
    publishPayload.release_time = Number(myClaimForUriEditing.timestamp);
  } else {
    publishPayload.release_time = Number(Math.round(Date.now() / 1000));  // new claim
  }
```
So, `release_time` will always be populated.

tzarebczan:
Gotcha...that makes more sense. Testing...

PR #414, opened 2021-06-04
Don't reset content language on edit

saltrafael:
## Issue

Publish page tries to set the upload language to the user langauge, and it does so on edit and overwrites the existing value.
https://github.com/lbryio/lbry-desktop/issues/6132

## Change

selectPublishFormValues was trying to return "language" from formValues but would always return false, and set the language as default instead, should return "languages" instead.
Also "pendingPublish" was unused


jessopb:
the bundle.js is built from the rest of the code. You would make this change in the publish selector if it is necessary.

saltrafael:
@jessopb you mean like this, or only on the publish selector?

PR #418, opened 2021-06-24
store pendingById

jessopb:
because we've been tracking pending using claim_list and it's hit or miss whether the confirmations=0 shows up on the first page or last page, persisting pending and checking them with txo_list will work better.


src/redux/actions/claims.js
```diff
@@ -18,14 +18,16 @@ import { creditsToString } from 'util/format-credits';
...
-import {
-  selectPendingCollections,
-} from 'redux/selectors/collections';
+import { selectPendingCollections } from 'redux/selectors/collections';
 import {
   doFetchItemsInCollection,
   doFetchItemsInCollections,
   doCollectionDelete,
 } from 'redux/actions/collections';
+import { selectPendingClaimsById } from '../selectors/claims';
```
infinite-persistence, line 27: Put into existing import?

infinite-persistence (dismissed):
- Not sure about the inline comments on merging claims too.
- [ ] While testing, I hit a scenario where a claim is stuck in "confirming..." state, and I don't see any further `txo_list` requests.  Clicking on the "Refresh" button didn't help.  Haven't tried F5, but I think that would fix it.
  - Can't recall my exact steps to reproduce, but it's something like this:
    - Create a post, and go to Uploads page to see the "confirming..." thing.
    - Quickly create another post.  
    - I now see 2 posts both "confirming..." (although the first post actually already confirmed.  I monitored this from another browser using the wallet page).
    - The first post shown as finished confirming within 30s later. The second post is stuck "confirming..." indefinitely.
- Previous discussions gave me the impression that `txo_list` is slow.  Looks ok during testing (300ms) -- I assume it's just to the simpler parameters.

jessopb:
Fixed the above bugs.

PR #419, opened 2021-07-05
Add claim in collections selector

saltrafael:
[Required for lbryio/lbry-desktop/pull/6393]

## Issue
https://github.com/lbryio/lbry-desktop/issues/6206 

## Change
Added selector to return if a given Claim is in any of a user's collection, needed because other selectors required a list ID

### DIFF
--- a/src/redux/selectors/collections.js
+++ b/src/redux/selectors/collections.js
@@ -166,19 +166,12 @@ export const makeSelectCollectionForId = (id: string) =>
 export const makeSelectClaimUrlInCollection = (url: string) =>
   createSelector(
     selectBuiltinCollections,
-    selectResolvedCollections,
+    selectMyPublishedCollections,
     selectMyUnpublishedCollections,
     selectMyEditedCollections,
     selectPendingCollections,
-    selectMyCollectionIds,
-    (bLists, rLists, uLists, eLists, pLists, myIds) => {
-      Object.entries(rLists).filter(
-        ([key, val]) =>
-          myIds.includes(key) &&
-          // $FlowFixMe
-          !pLists[key]
-      );
-      const collections = [bLists, uLists, eLists, rLists, pLists];
+    (bLists, myRLists, uLists, eLists, pLists) => {
+      const collections = [bLists, uLists, eLists, myRLists, pLists];
       const itemsInCollections = [];
       collections.map(list => {
         Object.entries(list).forEach(([key, value]) => {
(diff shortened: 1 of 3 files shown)


jessopb:
One thing about rLists (resolved) is it includes other peoples collections. So you'll have to filter rLists with myCollectionIds.

src/redux/selectors/collections.js
```diff
@@ -163,6 +163,27 @@ export const makeSelectCollectionForId = (id: string) =>
...
   );
 
+export const makeSelectClaimUrlInCollection = (url: string) =>
+  createSelector(
+    selectBuiltinCollections,
+    selectResolvedCollections,
+    selectMyUnpublishedCollections,
+    selectMyEditedCollections,
+    selectPendingCollections,
+    (bLists, rLists, uLists, eLists, pLists) => {
```
jessopb, line 173: rLists includes others collections resolved . have to filter with myCollectionIds

jessopb: see here https://github.com/lbryio/lbry-redux/blob/a327385cdf71568dbd15a17f3dcf5f4b83e0966d/src/redux/selectors/collections.js#L75

src/redux/selectors/collections.js
```diff
@@ -163,6 +163,34 @@ export const makeSelectCollectionForId = (id: string) =>
...
+export const makeSelectClaimUrlInCollection = (url: string) =>
+  createSelector(
+    selectBuiltinCollections,
+    selectResolvedCollections,
+    selectMyUnpublishedCollections,
+    selectMyEditedCollections,
+    selectPendingCollections,
+    selectMyCollectionIds,
+    (bLists, rLists, uLists, eLists, pLists, myIds) => {
+      Object.entries(rLists).filter(
```
jessopb, line 175: Here you're filtering an array made from rLists - not rLists itself so it seems the filtering is lost.

Since we're using this more than one place, we need to centralized a way to get myRLists with its own selector.

jessopb: https://github.com/lbryio/lbry-redux/blob/4cbb9a35c389dbe3962c54ee6d7844dcf25ae28e/src/redux/selectors/collections.js#L75
Maybe just using this list will save you some time and processing.

jessopb (approved):
This might be a bit faster with a different data structure for itemsInCollections, but I'll merge it.

